PR #5, opened 2024-10-29
Proposal section "Source routing"

falswe:
As discussed in class today, this is a proposal for the section "Source routing". It covers how to handle the flooding of the network and how to obtain the topology of it.


IlyaEmeliyanov (commented):
LGTM, can you please explain where the NodeId is coming from and how it's defined?

matteo-dinoia:
> LGTM, can you please explain where the NodeId is coming from and how it's defined?

I think we decided that we will be voting on this change next week

matteo-dinoia:
> > LGTM, can you please explain where the NodeId is coming from and how it's defined?
> 
> Until now we decided to use u64 for node ids, thanks for your comment.

Then either add a type definition (type NodeId =u64) in the documentation or convert to u64 (i prefer the first one because it is clearer in intent). Make sure it is used everywhere if you decide to introduce a type definition.

falswe:
> > > LGTM, can you please explain where the NodeId is coming from and how it's defined?
> > 
> > 
> > Until now we decided to use u64 for node ids, thanks for your comment.
> 
> Then please either add a type definition (type NodeId =u64) in the documentation or convert to u64 (i prefer the first one because it is clearer in intent). Make sure it is used everywhere if you decide to introduce a type definition.

There's no `NodeId` anymore in the document, I tried to use `u64` to be consistent with the rest of the document. I think it's not what this PR is primarily focusing on. Feel free to add it in another PR.

AP-protocol.md
```diff
@@ -62,26 +62,20 @@ As an example, consider the following simplified network:
...
-	/// Number of entries in the hops field.
-	/// Must be at least 1.
-	n_hops: usize;
-	/// List of nodes to which to forward the packet.
-	hops: [i64; 4],
-	/// Index of the receiving node in the hops field.
-	/// Ranges from 0 to n_hops - 1.
-	hop_index: u64,
+	/// Vector of nodes to which the packet will be forwarded to.
+	hops: Vec<u64>
```
StefanoSacchet, line 78: I would keep the hop_index just to avoid looping every time on the Vec in search of the id
```suggestion
	hops: Vec<u64>
	hop_index: u64
```

AP-protocol.md
```diff
@@ -102,22 +96,32 @@ struct Query {
...
-- Optionally, the node can return an **acknowledgment** or **path information** directly to the initiator, revealing its connection in the process.
+- If the query was not received earlier, the node forwards the updated message to its neighbours (except the one it received the query from) decreasing the TTL by 1, otherwise set the TTL to 0.
+- If the TTL of the message is 0, build a QueryResult and send it along the same path back to the initiator.
+
+```rust
+struct QueryResult {
+	flood_id: u64,
+	sourceRoutingHeader: SourceRoutingHeader,
+	path_trace: Vec<u64>,
+	node_types: Vec<NodeType>
```
StefanoSacchet, line 122:
```suggestion
	path_trace: Vec<(u64, NodeType)>
```

AP-protocol.md
falswe: I think the downside of this performance improvement is, that there could be a logical mismatch then, which would need to be dealt with. Like: What to do if the `hop_index` doesn't match the own position of `hops`?

AP-protocol.md
```diff
@@ -62,26 +62,20 @@ As an example, consider the following simplified network:
...
-	source_id: &'static str,
-	/// Number of entries in the hops field.
-	/// Must be at least 1.
-	n_hops: usize;
-	/// List of nodes to which to forward the packet.
-	hops: [i64; 4],
-	/// Index of the receiving node in the hops field.
-	/// Ranges from 0 to n_hops - 1.
-	hop_index: u64,
+	/// Vector of nodes to which the packet will be forwarded to.
```
matteo-dinoia, line 77:
```suggestion
	/// Vector of nodes to which the packet will be forwarded to.
	/// Contains itself as first element, and the destination as last element
```
Because it is not normal containing itself in the path we should specify this

AP-protocol.md
matteo-dinoia: I would make the drone just check the next hops. If the next hops is possible then it forward it, else it just drop it and sent the normal drop message.

matteo-dinoia (changes requested):

AP-protocol.md
matteo-dinoia: I would make the drone just check the next hops. If the next hops is possible then it forward it, else it just drop it and sent the normal drop message.

AP-protocol.md
matteo-dinoia: Yeah that is better maybe (we could eve do a small struct for it)

AP-protocol.md
falswe: It actually contains the initiator as the first element. I pushed a new commit to make this more clear.

matteo-dinoia (approved):
I think if we fix the other two minor thing that were brought up in this discussion, then it will be ready enough so that WG can vote.

AP-protocol.md
```diff
@@ -102,22 +96,31 @@ struct Query {
...
+	initiator_id: u64,
 	/// Time To Live, decremented at each hop to limit the query's lifespan.
+	/// When ttl reaches 0, we start a QueryResult message that reaches back to the initiator
 	ttl: u64,
 	/// Records the nodes that have been traversed (to track the connections).
-	path_trace: [u64; 20]
-	node_types: [NodeType; ]
+	path_trace: Vec<u64>;
+	// node_types is a vector with corresponding NodeTypes (NodeType may be an enum?)
+	node_types: Vec<NodeType>
```
matteo-dinoia, line 106:
```suggestion
	path_trace: Vec<(u64, NodeType)>
```
Should do here the same that it was done below

IlyaEmeliyanov (approved):

AP-protocol.md
```diff
@@ -102,22 +96,29 @@ struct Query {
...
-	initiator_id: RC<Arc<usize>>,
+	initiator_id: u64,
 	/// Time To Live, decremented at each hop to limit the query's lifespan.
+	/// When ttl reaches 0, we start a QueryResult message that reaches back to the initiator
 	ttl: u64,
 	/// Records the nodes that have been traversed (to track the connections).
-	path_trace: [u64; 20]
-	node_types: [NodeType; ]
+	path_trace: Vec<(u64, NodeType)>
 }
```
matteo-dinoia, line 105:
```suggestion
}

enum NodeType{Client, Drone, Server}
```

matteo-dinoia:
We have a big problem with this PR. "It is NOT POSSIBLE to use Vec in anything that will go inside a fragment", and because sourceRouting NEED to go inside it, we cannot use Vec.

AP-protocol.md
DaYe03: we could make NodeId reconizable like every drones id starts with D (D1 D2) every clients starts with C and server with C.

matteo-dinoia: For Node Id propose please refer to #19. In here anything is fine (it's just a temp Type).

falswe: Done

falswe:
> We have a big problem with this PR. "It is NOT POSSIBLE to use Vec in anything that will go inside a fragment", and because sourceRouting NEED to go inside it, we cannot use Vec.
> 
> I propose using a long array [N, NodeId] terminated with -1.

Can't this be considered in the fragmentation process? I would say the vector here should be fine (like also the professor said in the last meeting), but in the produced fragments there can't be a vector anymore. This could be handled by the fragmentation process.

matteo-dinoia:
> > We have a big problem with this PR. "It is NOT POSSIBLE to use Vec in anything that will go inside a fragment", and because sourceRouting NEED to go inside it, we cannot use Vec.
> > 
> > I propose using a long array [N, NodeId] terminated with -1.
> 
> Can't this be considered in the fragmentation process? I would say the vector here should be fine (like also the professor said in the last meeting), but in the produced fragments there can't be a vector anymore. This could be handled by the fragmentation process.

No it cannot be done, because the entire vec of hops must fit inside the packets fixed size, or else part of the path is not inside the packets so it cannot reach the destination. 
So please change the vec to a large array or other solution.

matteo-dinoia:
> > We have a big problem with this PR. "It is NOT POSSIBLE to use Vec in anything that will go inside a fragment", and because sourceRouting NEED to go inside it, we cannot use Vec.
> > I propose using a long array [N, NodeId] terminated with -1.
> 
> Can't this be considered in the fragmentation process? I would say the vector here should be fine (like also the professor said in the last meeting), but in the produced fragments there can't be a vector anymore. This could be handled by the fragmentation process.

I stand corrected. I asked the professor. Message Fragment must be of fixed size, but headers and everything else can be of not-fixed size, so this is fine. (I'll update the protocol to reflect better this limitation).

AP-protocol.md
```diff
@@ -130,7 +133,8 @@ For every response or acknowledgment the initiator receives, it updates its unde
 
 The flood can terminate when:
 
--
+- Timeout after a certain period, depending on the number of nodes and on the maximum timeout in communication channels
```
luiss07, line 136: We think this is a Client behaviour. The termination condition should refer to the condition that triggers the send of the QueryResult message and for us the TTL is enough.

AP-protocol.md
```diff
@@ -130,7 +133,8 @@ For every response or acknowledgment the initiator receives, it updates its unde
...
+- A drone crashes in the network discovery phase. As soon as the initiator knows it, it starts a new flooding with a new flood_id, and it starts ignoring QueryResult messages about the previous flood_id.
```
luiss07, line 137: We think this is a Client behaviour. The termination condition should refer to the condition that triggers the send of the QueryResult message and for us the TTL is enough.

luiss07, line 137: We think this is a Client behaviour. The termination condition should refer to the condition that triggers the send of the QueryResult message and for us the TTL is enough.

AP-protocol.md
```diff
@@ -98,26 +92,35 @@ To do so, they must use the **Network Discovery Protocol**. The Network Discover
...
 ```
 
 ### **Neighbor Response**
 
 When a neighbor node receives the query, it processes it based on the following rules:
 
-- If the query was received earlier (i.e., the query identifier is already known), the node broadcasts the query to increase the robustness.
-- Otherwise, the node appends its own ID and type (that is, Drone, Client, Server) to the **path trace** and forwards the query to its neighbors (except the one it received the query from).
-- Optionally, the node can return an **acknowledgment** or **path information** directly to the initiator, revealing its connection in the process.
+- If the query was not received earlier, the node forwards the updated message to its neighbours (except the one it received the query from) decreasing the TTL by 1, otherwise set the TTL to 0.
```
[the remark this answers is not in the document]
brokenhouse: We could then take advantage of **received earlier** as a condition for which its possible to build a `QueryResult` (instead of waiting every time for the ttl to become 0) as it would mean the protocol has discovered a cycle in the network and it's not possible to gather further information about the topology.

AP-protocol.md
matteo-dinoia, line 137:
```suggestion

```

PR #10, opened 2024-10-30
TTL update & remove struct AckInner

### LINKED ISSUE #9: Update TTL type & remove struct AckInner
### TTL
Just wanted to suggest to limit the size of the **ttl** property in all of the structs.
There's no need to have it that big. A `u8` type might be sufficient, it'll save space and it'll align with the network topology and the modern internet standards with a max **TTL** of 255. Also a possible default value could be 64.

### `struct AckInner {}`
In the definition of the Ack struct we're using the AckInner struct like this:
```
pub struct Ack(AckInner);

struct AckInner {
	session_id: u64,
	when: std::time::Instant,
	// Time at which the message was received.
}
```
I think it's redundant and I suggest to remove AckInner and move the attributes in the Ack struct itself. Like this:
```
struct Ack {
	session_id: u64,
	when: std::time::Instant,
	// Time at which the message was received.
}
```

IlyaEmeliyanov:
Resolves #9

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -105,7 +105,7 @@ struct Query {
 	/// ID of client or server
 	initiator_id: RC<Arc<usize>>,
 	/// Time To Live, decremented at each hop to limit the query's lifespan.
-	ttl: u64,
+	ttl: u8,
 	/// Records the nodes that have been traversed (to track the connections).
 	path_trace: [u64; 20]
 	node_types: [NodeType; ]
@@ -262,7 +262,7 @@ struct Error {
 	session_id: u64,
 	/// ID of drone, server of client that is not a neighbor:
 	id_not_neighbor: String,
-	ttl: u32,
+	ttl: u8,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


AP-protocol.md
```diff
@@ -436,4 +434,4 @@ Notice that these messages are not subject to the rules of fragmentation, in fac
 - S -> C : client_list!(list_length, list_of_client_ids)
 - C -> S : message_for?(client_id, message_size, message)
 - S -> C : message_from!(client_id, message_size, message)
-- S -> C : error_wrong_client_id!
\ No newline at end of file
+- S -> C : error_wrong_client_id!
```
matteo-dinoia, line 437: Not sure what changed here (maybe you changed the spacing or removed/added \n)

AP-protocol.md
```diff
@@ -287,9 +287,7 @@ Source Routing Header contains the path to the client, which can be obtained by
 If a drone receives a Message and can forward it to the next hop, it also sends an Ack to the client.
 
 ```rust
-pub struct Ack(AckInner);
-
-struct AckInner {
+struct Ack {
```
matteo-dinoia, line 290: This is fine but maybe it should be in another pull request.
At least it should be in another commit.

And same for the pub in the L336

matteo-dinoia (changes requested):
Seem fine by me only fix the problem cited

AP-protocol.md
IlyaEmeliyanov: Ok

matteo-dinoia:
Can you rebase on newer changes (or just create a new PR) with only the u8 stuff? Then I'll merge it if nobody is against it.

PR #11, opened 2024-11-02
Proposal to change Nework initialization file format to TOML

marcoff181:
We propose changing the `Network Initialization File` format to a `.toml` file with the advantages of:
- having the ability to add a variable amount of nodes of each type (drone/client/server), which could help with debugging
- making it easier to add other parameters in the future if needed
- no need to spend time writing a parser, the code example below  is all that's needed to parse from the proposed config file:
```rust
use serde::Deserialize;  
use std::fs;  
  
#[derive(Debug, Deserialize)]  
struct Drone {  
    id: u64,  
    connected_drone_ids: Vec<u64>,  
    pdr: f64,  
}  
  
#[derive(Debug, Deserialize)]  
struct Client {  
    id: u64,  
    connected_drone_ids: Vec<u64>,  
}  
  
#[derive(Debug, Deserialize)]  
struct Server {  
    id: u64,  
    connected_drone_ids: Vec<u64>,  
}  
  
#[derive(Debug, Deserialize)]  
struct Config {  
    drone: Vec<Drone>,  
    client: Vec<Client>,  
    server: Vec<Server>,  
}  
  
fn main() {  
    let config_data = fs::read_to_string("src/config.toml").expect("Unable to read config file");  
    let config: Config = toml::de::from_str(&config_data).expect("Unable to parse TOML");  
    println!("{:#?}", config);  
}
```

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -48,31 +48,31 @@ connected_drone_ids = ["connected_id1", "connected_id2", "connected_id3", "..."]
 ### Example
 ```toml
 [[drone]]  
-id = "d1"  
-connected_drone_ids = ["d3", "d4", "d5"]  
+id = 1  
+connected_drone_ids = [2, 3]  
 pdr = 0.05  
   
 [[drone]]  
-id = "d2"  
-connected_drone_ids = ["d1", "d8"]  
+id = 2 
+connected_drone_ids = [1,3,4]  
 pdr = 0.03  
   
 [[drone]]  
-id = "d3"  
-connected_drone_ids = ["d2", "d6", "d5"]  
+id = 3  
+connected_drone_ids = [2,1,4]  
 pdr = 0.14  
   
 [[client]]  
-id = "c1"  
-connected_drone_ids = ["d2", "d3"]  
+id = 4 
+connected_drone_ids = [3, 2]  
   
 [[client]]  
-id = "c2"  
-connected_drone_ids = ["d1"]  
+id = 5  
+connected_drone_ids = [1]  
   
 [[server]]  
-id = "s1"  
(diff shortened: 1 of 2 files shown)


AP-protocol.md
```diff
@@ -3,38 +3,77 @@
...
+connected_drone_ids = ["connected_id1", "connected_id2", "connected_id3", "..."] # at least 2 entries
+```
+- note that `connected_drone_ids` cannot contain `server_id` nor repetitions
+- note that a server cannot connect to other clients or servers
+- note that a server should be connected to at least two drones
+
+### Example
+```toml
+[[drone]]  
+id = "d1"
```
matteo-dinoia, line 51: The only thing that should be changed is that we are using elsewhere u64 for client and server (and having them as a string is stupid)

matteo-dinoia (changes requested):

matteo-dinoia:
Also because your example may be useful for someone who never used this crate, i would like you to create a file rust (in a folder like examples) and add the code and may how to import it as a comment.
``` Toml
[dependencies]
toml = "0.8.19"
serde = { version = "1.0.214", features = ["derive"] }
```

That is what I think at least. If you think it is not necessary let me know.

AP-protocol.md
marcoff181: Thanks, I didn't notice I had left them in as strings, fixed in df61622

matteo-dinoia (approved):
Ok, i think it is ready for vote

matteo-dinoia:
mmmm next time pleas use a normal name for the branch. It is so long it is causing problem

matteo-dinoia:
Merged manually because i have no access to your pull request.

PR #12, opened 2024-11-02
Simplify Message and Packets (+ adding NodeId)

matteo-dinoia:
In this PR:
1. Message type is complex. To simplify it I put request and response first and foremost (because server is only interested in request and send only response).
2. I also merged Text and Media Request Response (the enums)
3. added serverType (could be better)
4. added NodeId for readability
5. added example of use of Message Content

Edit: see comment below for newest changes


Marui-9:
Makes sense 👍

matteo-dinoia:
Set as priority because it create base for common code

matteo-dinoia:
Update code to remove the high level of nesting in favor of even simpler type.
Testing the usability of the nested code,  I found that while it did add a lot of complexity during creation, it didn't simply almost anything in unpacking.

IlyaEmeliyanov (changes requested):
Well done, **small change**: I think that for the **ServerType** it'd be cleaner to use an **enum** with 3 variants (Chat, Text and Media) instead of a struct with booleans.

matteo-dinoia:
> Well done, **small change**: I think that for the **ServerType** it'd be cleaner to use an **enum** with 3 variants (Chat, Text and Media) instead of a struct with booleans.

Yes, but it wouldn't allow for server of multiple kind which the professor was in favor of. I think we can vote on this separately.

Suge42:
The changes makes sense, but the simplification of the Message type is reverting a previous changed on which all WGL agreed, so we'll vote about that.
Other than that, the rest of the changes like Nack, those at packets and similar, and other small things looks fine.

AP-protocol.md
```diff
@@ -157,141 +166,87 @@ Message is subject to fragmentation: see the dedicated section.
...
-}
-
-struct MediaRequest{
-	Media{  //C -> S : media?(media_id, +media type+ )
-		media_id: u64
-		///media: media_kind
-	}
+#[derive(Debug)]
+pub struct MessageData { // Only part fragmentized
+	source_id: NodeId,
```
luiss07, line 185:
```suggestion
```
The `source_id` is already present into the `SourceRoutingHeader`.

matteo-dinoia: Fair. I'll remove it now. i still think it is still may be better to keep it here to isolate different layer. But i'll open a PR about it.

AP-protocol.md
```diff
@@ -157,141 +166,86 @@ Message is subject to fragmentation: see the dedicated section.
...
 ```rust
-struct Error {
-	session_id: u64,
-	/// ID of drone, server of client that is not a neighbor:
-	id_not_neighbor: String,
-	ttl: u32,
-}
+let routing = getRoutingHeader();
+let content = MessageType:ReqFile(8);
+Message::new(ruting, source_id, session_id, content)
```
brokenhouse, line 222:
```suggestion
Message::new(routing, source_id, session_id, content)
```

AP-protocol.md
```diff
@@ -157,141 +166,86 @@ Message is subject to fragmentation: see the dedicated section.
...
 
-This message cannot be dropped by drones due to Packet Drop Rate.
+This message cannot be dropped by drones due to Packet Drop Probability.
 
 ```rust
-struct Dropped {
-	session_id: u64,
+pub enum Nack{
+	ErrorInRouting(NodeId), // contains id of not neighbor
+	Dropped
```
chiarastrozzii, line 238: i think it would be useful to keep track of which fragment we have dropped, just to know which one we have to send again

Marui-9:
The `Message(Fragment)` naming is ambiguous as it could introduce a conflict between low level fragments and higher level Messages. It could be be beneficial to have something like `FragPack(Fragment)`.

PR #15, opened 2024-11-02
Adding basic starting action

matteo-dinoia:
This is a basic action, created from the one presented during lesson.
Please give suggestion.


matteo-dinoia:
Merging it because is better than nothing

PR #17, opened 2024-11-03
Adding basic API and Topology

Marui-9:


IlyaEmeliyanov:
Good starting point 👍🏻

src/types/topology/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+pub struct Topology {
+    nodes: Vec<NodeRef>,
+}
+
+pub struct Node {
+    name: NodeId,
+    node_type: NodeType,
+    neighbors: HashMap<NodeId, NodeRef>, //node ids
+}
+pub enum NodeType {
```
Suge42, line 18: Isn't it repetitive to have NodeId as 'name' in Node, and also in NodeType?

Marui-9: You're right

Suge42: Other than that seems like a good start 👍

src/api/drone.rs
```diff
@@ -0,0 +1,23 @@
...
+    thread::spawn(move || {
+        let _id = drone_id;
+        let receivers = chnl_recv;
+        let _senders = chnl_send;
+
+        for i in &receivers {
+            i.recv().unwrap();
+        } //This would listen to all the channel for which this drone is a receiver
+    });
+}
```
matteo-dinoia, line 19: I'm very against the idea to add code like this to the common code. This is code that should be defined by the group not by the class. 

We should focus on struct, API and the like. This code in particular touch a portion of the code-base that probably every group would want to do in different way.

src/types/message.rs
```diff
@@ -0,0 +1,29 @@
...
+struct Message {}
+pub struct Fragment {}
+impl Message {
+    #[allow(unused_variables)]
+    fn serialize(&self) -> String {
+        unimplemented!()
+    }
+    //takes message and returns the data struct serialized in a String
+    //so it goes from the actual data struct to a String
+    #[allow(unused_variables)]
```
matteo-dinoia, line 11: You probably need to `run cargo fmt --all`

src/types/message.rs
```diff
@@ -0,0 +1,29 @@
+//use crate::types::packet::Fragment;
+struct Message {}
+pub struct Fragment {}
+impl Message {
+    #[allow(unused_variables)]
+    fn serialize(&self) -> String {
+        unimplemented!()
+    }
```
matteo-dinoia, line 8: Having a struct with methods unimplemented does not seem right and this code. Because this code should be implemented by groups not class, we maybe should use a trait instead

src/types/topology/routes.rs
```diff
@@ -0,0 +1,10 @@
+use crate::types::topology::{NodeId, NodeRef};
+use std::collections::HashMap;
+
+pub struct Route {
+    //route is the actual vec of references, path is just a "treasure map" with ids
+    path: Vec<NodeRef>,
+}
```
matteo-dinoia, line 7:
```suggestion
type Route = Vec<NodeRef>;
```

src/types/topology/routes.rs
```diff
@@ -0,0 +1,10 @@
...
+pub type Routes = HashMap<NodeId, Route>;
```
matteo-dinoia, line 8:
```suggestion
```
Why we have an hashmap of routes wouldn't it be better using a vec? Also I don't see the gains in using a type in this case i would just remove it. Expecially because it can cause confusion.

src/types/topology/routes.rs
```diff
@@ -0,0 +1,10 @@
...
+//Routes are only from client to server and vice versa.
```
matteo-dinoia, line 9: Unneeded comment in my opinion. If you want to add a constraint you should propose that and add it to the protocol.

src/api/drone.rs
Suge42: Yea I asked him to add it, but probably tomorrow we'll go back, because it was just a sketch and shouldn't have put it there already

Suge42: But since all the drones will need to be compatible with the same initializer, at least a common idea on what to do is needed.

matteo-dinoia: Ok i understand what you are coming from. Than that is fine but it shouldn't be in this PR. Also i think that ypu just need to create the struct the initializer should be the one externally to create a thread and run this drone on it.

src/api/drone.rs
```diff
@@ -0,0 +1,6 @@
+use crate::types::topology::routes::Path;
+
+pub struct Packet {}
+pub trait DroneAble {
+    fn forward_packet(&self, path: &Path, hop: usize, packet: Packet) -> bool;
```
Paglia20, line 5: is the path field necessary? isn't it already comprehended in the source header part of the packet field?

PR #19, opened 2024-11-03
New Id Format

panigiove:
This proposal introduces a new ID format for clients, servers, and drones within the Network Initialization File. These changes aim to make it easier for the Simulation Controller to identify node information, including a drone's owner and a node's type, and also for debugging purposes.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -8,6 +8,14 @@ This document also establishes some technical requirements of the project.
 
 The **Network Initializer** reads a local **Network Initialization File** that encodes the network topology and the drone parameters and, accordingly, spawns the node threads and sets up the Rust channels for communicating between nodes.
 
+## Id Format
+
+ Each ID consists of three fields in the following format: ABN.
+
+- **A** represents the type of node: `0` for client, `1` for server, and `2` for drone.
+- **B** indicates the group that created the node (it is always `00` for clients or servers).
+- **N** represents the node number.
+
 The Network Initialization File has the following format:
 
 - 10 drone lines: `drone_id connected_drone_count connected_drone_ids packet_drop_rate`.
@@ -33,6 +41,8 @@ The Network Initialization File has the following format:
     Note that `connected_drone_ids` cannot contain `server_id` nor repetitions.
 
     Note that a server cannot connect to other clients or servers.
+
+Every `drone_id`, `client_id` and `server_id` respect the Id Format defined above. 
 
 Importantly, the Network Initializer should also setup the Rust channels between the nodes and the Simulation Controller (see the Simulation Controller section).
 


AP-protocol.md
```diff
@@ -8,6 +8,14 @@ This document also establishes some technical requirements of the project.
 
 The **Network Initializer** reads a local **Network Initialization File** that encodes the network topology and the drone parameters and, accordingly, spawns the node threads and sets up the Rust channels for communicating between nodes.
 
+## Id Format
+
+ Each ID consists of three fields in the following format: ABN.
+
+- **A** represents the type of node: `0` for client, `1` for server, and `2` for drone.
+- **B** indicates the group that created the node (it is always `00` for clients or servers).
```
matteo-dinoia, line 16: Do we need B? And what is B for drones?

AP-protocol.md
```diff
@@ -8,6 +8,14 @@ This document also establishes some technical requirements of the project.
...
+- **N** represents the node number.
```
matteo-dinoia, line 17: Either N is left padded by 0 or we treats ABN as a String or it will break the format.

Counter proposal: format NA

marcoff181:
Just to note: if the WG approves PR #11 this approach becomes impractical, but the idea of adding a way to know which group implemented which drone is still useful. 
This PR could then become "adding info about drone implementation" and solve the issue by adding a field to drone in the new format of the config file

AP-protocol.md
DaYe03: The purpose of the field `B` is to simplify the identification of the specific group responsible for building each drone. This has several advantages:

1. **Easier Scenario Setup**: By identifying groups through field `B`, it's quicker to organize and simulate specific scenarios.

2. **Streamlined Drone Creation**: Field `B` allows for straightforward creation of new drones associated with a particular group, as they can be assigned based on the information stored in this field.

3. **Enhanced UI Display**: Having this grouping information makes it easier to display detailed information about each drone on the user interface.

Overall, `B` provides a practical way to manage and organize drones efficiently

DaYe03:
> Just to note: if the WG approves PR #11 this approach becomes impractical, but the idea of adding a way to know which group implemented which drone is still useful. This PR could then become "adding info about drone implementation" and solve the issue by adding a field to drone in the new format of the config file

Yes, the alternative would be to separate the drone's group ID as an attribute of the drone itself. However, I don’t think it’s impractical: because it would still be in the ID field of individual components (server, client, drones). This would actually shorten the path vector in flooding packets and messages, as clients wouldn’t need additional information to understand which component an ID belongs to. The UI or initializer could use this ID alone to determine which drone to create and which group it belongs to.

matteo-dinoia:
> ding a field to drone in the new format of the config file

Why is unpractical? we just get the ID and create function from ID to partial info (with just the equivalent of a bit mask and bit shift).

AP-protocol.md
```diff
@@ -8,6 +8,14 @@ This document also establishes some technical requirements of the project.
 
 The **Network Initializer** reads a local **Network Initialization File** that encodes the network topology and the drone parameters and, accordingly, spawns the node threads and sets up the Rust channels for communicating between nodes.
 
+## Id Format
+
+ Each ID consists of three fields in the following format: ABN.
```
matteo-dinoia, line 13: I would advice at using the inverse NBA:
so we have n, 2, 1 bit.

For example if we want to be able to support big number of drone that is possible with this methods without having to do something like.
"ABBNNNNNNNNNNNNNNNNNNNNNN" that would make it very long and need to require a fixed lengthof the N.

Not sure if it is clear.
If we do so than I highly recommend using 1 as the smaller id possible.

marcoff181:
> Why is unpractical? we just get the ID and create function from ID to partial info (with just the equivalent of a bit mask and bit shift).

Sorry, unpractical was a bad way ot saying this. To me it just seemed more clear if the drone ID and group ID were two separate fields

matteo-dinoia:
> > Why is unpractical? we just get the ID and create function from ID to partial info (with just the equivalent of a bit mask and bit shift).
> 
> Sorry, unpractical was a bad way of saying this. To me it just seemed more clear if the drone ID and group ID were two separate fields

Ok, i think it is fine in a single field, else it will become a mess to reference other nodes. A single integer is for the best.

AP-protocol.md
brokenhouse, line 16: naming them T for the type and G for group would be more meaningful

DaYe03: yeah we will change that

matteo-dinoia:
Voted against

PR #24, opened 2024-11-05
Change chat server entry in Message Content

matteo-dinoia:
In this PR there are two changes:
1. ability for client to register to a chat server (it is needed because else the server doesn't know the clients)
2. ability and requirement for the client to fetch data instead of the server sending them data immediately when new data is available

### DIFF
--- a/src/types/message.rs
+++ b/src/types/message.rs
@@ -35,6 +35,8 @@ pub enum MessageContent {
 
     ReqClientList,
     ReqRegisterInClientList,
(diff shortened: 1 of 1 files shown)


matteo-dinoia:
TODO: it is necessary to use registration, and last part can change

matteo-dinoia:
TODO: change in the protocol file also

matteo-dinoia:
Removed as it is unneeded and we have more important things to do.

PR #26, opened 2024-11-06
Update of drone API

Suge42:
First try to a drone API considering threads and crossbeam channels


matteo-dinoia:
This pull request should only contains channel by what I see, other part are either part of other PR (see #23) or not relevant and should be moved to other PR.

PS. I'll fully review it later.

PR #27, opened 2024-11-07
Drone api

Suge42:


src/api/drone.rs
```diff
@@ -1,5 +1,20 @@
...
+use crate::types::channel::Channel;
+use crate::types::source_routing_header::NodeId;
+use std::collections::HashMap;
+use std::thread;
 
-pub trait DroneAble {
-    fn forward_packet(&self, packet: Packet) -> bool;
+pub struct Drone {
+    id: NodeId,
+    thread: thread::JoinHandle<()>,
```
matteo-dinoia, line 8: Not sure if thread should be inside of it or not. I'm not perfectly up to speed with thread in rust but it seem to me to have the drone not think about being a thread.

src/api/drone.rs
```diff
@@ -1,5 +1,20 @@
...
+}
+
+trait DroneTrait {
+    fn new(drone_id: u8) -> Self;
+    // The thread would be created inside here,
+    // giving him reference to the list of channels,
+    // that would be expanded during the creation of the
+    // later drones.
+    fn add_channel(new_channel: Channel) -> Result<String, String>;
+    fn remove_channel(drone_id: u8) -> Result<String, String>;
```
matteo-dinoia, line 19: Result<String, String> seem excessive to me. Either Result<(), String> or Result<id/size, String> or bool seem better.

matteo-dinoia (commented):
While the file channel.rs and drone.rs (+ cargo.toml) are fine. The other should be in another PR so that i can merge them earlier or at least so that they are clearly divided. If you need help with that just msg me.

matteo-dinoia:
Also while it is possible for you to create branch on this repository, please don't do this. Fork the repository and make pull request from there.

src/api/drone.rs
Suge42: The idea was that while the thread needs to have at its disposal the channels, you also need to store them elsewhere for where you want to add a new drone with the Simulation Controller.
Having a struct Drone like this allows to keep track of the list of channels of each drone in an ordered way, and also save a reference to the drone (that might be needed for the crash, but I'm not sure about how do to that still).

src/api/drone.rs
Suge42: Yea it was more a placeholder than anything.

Suge42:
> While the file channel.rs and drone.rs (+ cargo.toml) are fine. The other should be in another PR so that i can merge them earlier or at least so that they are clearly divided. If you need help with that just msg me.

About the other files, the small changes are made to adjust some wrong allocation of variable in the previous proposal, so unless we separated them and pulled the change, I'm not sure it would be time efficient to do another pull request

matteo-dinoia:
It's not only about time (which requires little to no time) but also to readability during the vote in class. And also merging small stuff like this already reduce the probability of conflict in the future, making it easier to work for people.

PR #33, opened 2024-11-12
Updating packets (adding query and removing fragment data)

matteo-dinoia:
This PR also add the id on file/media response.

All code in this PR is from @ZGako.


IlyaEmeliyanov:
seems good to me

matteo-dinoia:
I have removed the part about adding ID to the response (please open a new PR about it), merging the rest.

PR #36, opened 2024-11-12
DroneAPI

Suge42:
New iteration of the drone API, that combines the two proposal shown in the last WG.

The structure proposed aims to be created inside a thread, where it'll contain all the information needed to running the drone.
The list of the channel on which the drone can communicate will be updated a runtime via a command from the channel that will also be assigned to the simulation controller (might also be used during creation by the iniziator, since the senders can be duplicated).

The struct Drone is now strictly API, but we'll still leave the example of implementation availbe, for those who might need it.

### DIFF
--- a/examples/drone_usage_example.rs
+++ b/examples/drone_usage_example.rs
@@ -6,7 +6,14 @@ use crossbeam_channel::{select, Receiver, Sender};
 use std::any::Any;
 use std::collections::HashMap;
 
-fn main() {}
+fn main() {
+    //Create some drone
+    // let drone = Drone::new(...);
+
+    //Then start the drone
+    //(not forced to use '.run()', a loop is enough).
+    // drone.run();
+}
 
 impl DroneImplement for Drone {
     fn new(
--- a/examples/drone_usage_example.rs
+++ b/examples/drone_usage_example.rs
@@ -1,9 +1,9 @@
 //THIS IS JUST AN EXAMPLE OF IMPLEMENTATION
 
 use crate::api::drone::Drone;
+use crate::types::command::Command;
 use crate::types::packet::{Packet, PacketType};
 use crate::types::source_routing_header::NodeId;
-use crate::types::command::Command;
 use crossbeam_channel::{select, Receiver, Sender};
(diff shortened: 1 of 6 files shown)


src/api/drone.rs
```diff
@@ -1,5 +1,26 @@
...
+pub struct Drone {
+    pub drone_id: NodeId,
+    pub sim_contr_send: Sender<Command>,   //Not packet.
+    pub sim_contr_recv: Receiver<Command>, //Not packet.
+    pub packet_send: HashMap<NodeId, Sender<Packet>>, //All the sender to other nodes.
+    pub packet_recv: Receiver<Packet>, //This drone receiver, that will be linked to a sender given to every other drone.
+    pub pdr: u8,                       //Would keep it in % to occupy less space, but could be f32.
+}
+
+pub trait DroneImplement {
```
matteo-dinoia, line 17: DroneImplement is not a descriptive name

examples/drone_usage_example.rs
```diff
@@ -0,0 +1,66 @@
...
+use std::thread;
+
+fn main() {
+    let handler = thread::spawn(move || {
+        //Create some drone
+        // let drone = Drone::new(...);
+
+        //Then start the drone
+        //(not forced to use '.run()', a loop is enough).
+        // drone.run();
```
matteo-dinoia, line 17: run should be part of DroneImplement so that simulation controller can threat them as opaque. Without needing to knoe how to start each single drone.

matteo-dinoia (commented):
I don't like the split between drone and drone implement. I'll test some things when I have time and see if I find a better way.

matteo-dinoia:
Made a PR to your PR [Here #1](https://github.com/Suge42/WGL_repo_2024/pull/1)

examples/drone_usage_example.rs
Suge42: You're right, I'll move it.

src/api/drone.rs
Suge42: I know, but I didn't have a better idea.

PR #38, opened 2024-11-16
Rules that the Drone must follow

matteo-dinoia:
Create standard rules for what the drone must do. They are very simple and are only there to avoid drone with non standard behavior.

Note: Still missing, the client should also respond in case it receive a package from a client (which is not normal). Should it respond or just silently drop it?

### DIFF
--- a/src/types/packet.rs
+++ b/src/types/packet.rs
@@ -19,9 +19,10 @@ pub struct Nack {
     nack_type: NackType,
 }
 
-pub enum NackType {
-    ErrorInRouting(NodeId), // contains id of not neighbor
-    Dropped(),
+pub enum NackType{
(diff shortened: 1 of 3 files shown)


src/types/packet.rs
```diff
@@ -21,7 +21,8 @@ pub struct Nack {
 
 pub enum NackType {
     ErrorInRouting(NodeId), // contains id of not neighbor
-    Dropped(),
+    DestinationIsNotAServer,
```
Suge42, line 24: I think that this should be called 'DestinationIsDrone' as the other one, since clients are also possibile destination. Other than that seems good to me, we'll vote on this tomorrow.

matteo-dinoia: Fixed

matteo-dinoia:
TODO: add back index everywhere

PR #40, opened 2024-11-18
From executable to library

DiCesareDaniele:
# Introduction
Sorry in advanced (haha got it?) for the late request, I know that reviewing PR right before the WG is not enjoyable but with the exam and everything I delayed. That said, I think we can postpone this for a bit as it will also break every open PR. 

The purpose is to be able to add this repo as a dependency in each group's Cargo.toml as now it's not possible cause the project is structured as an executable and not as a library. With this we'll be able to easily do that.

_Cargo.toml_
```toml
[dependencies]
wg_2024 = { git = "https://github.com/WGL-2024/WGL_repo_2024.git" }
# if you want to test before this PR get's approved substitute the git url with that of the forking repo
```
_drone.rs example_
```rust
use wg_2024::drone::{Drone, DroneOptions};

struct MyDrone;

impl Drone for MyDrone {
    fn new(_: DroneOptions) -> Self {
        todo!()
    }
    fn run(&mut self) {
        todo!()
    }
}
```

**⚠️ Note: no code was harmed during this refactoring ⚠️**

Except for the following
- some code in example got fixed 🔨
- network topology config structs were promoted 👍, good for them 😀
- drone new method now takes a DroneOptions struct instead of the 69420 parameters it took before 👎

# About the project structure
Every package is located in the crates folder, so in this section I will assume that as our working directory

- **wg_config**: the structs for parsing the network topology file
- **wg_controller**: everything related to the simulation controller (e.g. Command passed to the drones)
- **wg_drone**: the drone api
- ~**wg_network**: everything regarding the topology~ _this section will probably get removed in the future_
- **wg_packet**: the Message and Packets structs

# Why workspaces
**⚠️ Note: this section does not contain useful information, it's just for clarification and usage if you want to skip this feel free to do so, just make sure to open all the links first ⚠️**

After talking with @matteo-dinoia and other members a problem arised... the much feared workspace 😧. 
I wanted to use workspaces cause from my experience they keep library code cleaner and provide more modularity and separation of concern than just using modules, as a bonus they are really easy to use and understand. But rightfully so not everyone wants to have yet another thing to learn. The good thing is that you don't have to [link](https://www.youtube.com/watch?v=lrprCTvgI8c&t=38s), using a library with workspaces is the same as using one without. The only place where it can make a difference is in the repo code itself, even then it's the same thing as just using modules but if you want to import a package from another package you have to specify it like this in the package Cargo.toml
```toml
name_of_package = { path = "../name_of_package" }
```
just be careful not to create a circular dependency [link](https://www.youtube.com/watch?v=m1zhuwhrmwA&t=19s)
and if you want to have another package just run the following in the root of the project and add the package to crate/wg_internal (for exporting)
```bash
cargo new crates/name_of_package --lib
```
That said if we find this to be too challenging I guess we can stick to regular modules [link](https://www.youtube.com/watch?v=OZIneMCkX_4&t=59s)

I have lost more time on writing this description than on writing the code itself, so with this I am out [link](https://www.youtube.com/watch?v=kBW37Gqq_m4&t=27s).


Marui-9:
Magnificent

matteo-dinoia:
The code work. Only three notes:

1. we should add how to import this repository to the `README.md`
2. Putting the example inside the library make it needed to import `serde` and `deserialize` even if some group may not use it. Also this kind of example make little sense to stay in the library itself. 
    * Maybe we should move the examples to their own repo that import this one
3. While I agree with you that subcrate (or whatever they are called) are useful for keeping the code cleaner, I'm not if this will slow down the already slow contribution rate. This is especially important because in a week we will have to freeze the specification.

For the rest this PR is good. Next time try to keep random fixing and file moving in two different PR, because like this is difficult to see what changed.

matteo-dinoia:
About 3 for now leave it like it is now (with subcrates).

Ps. in 3 i meant "i'm not sure if it is okay"

PR #44, opened 2024-11-19
Dropping a query if it was already delivered

Seregno:
I think that the drone should drop the query if its NodeId has already been inserted in the path_trace vector. The benefits deriving from this restrain towards the drones are the following:
 - Making the implementation of the drone more easy since we won't have to remove the cicle form the path_trace or delegate this task to a client.
- There will be less queries "traveling" in the network during the network descovery phase.
- As a consequence of the previous point, the client will have to manage less answers from the nodes, focusing only on the paths 
that can potentially lead to a Server node.
- Minimize the probability of a client sending a message that has a cicle in its path. 
- The network discovery implementation can focus on reaching a server from a client rather than managing cicles among the nodes.

Point 3 and 5 focuses mainly on a topic that is up to the group rather than the protocol itself, but I think that this means my idea will simplify the implementation of the other components of this project in the future.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -59,6 +59,8 @@ A drone is characterized by a parameter that regulates what to do when a packet 
 
 Packet Drop Rate: The drone drops the received packet with probability equal to the Packet Drop Rate.
 
+The PDR can be up to 100%, and the routing algorithm of every group should find a way to eventually work around this.
+
 # Messages and fragments
 
 Recall that there are: Content servers (that is, Text and Media servers) and Communication servers. These servers are used by clients to implement applications.


matteo-dinoia:
I'm pretty sure that adding this change would let you lose some edges. With this methods you can at max discover N edges where N is the number of nodes. But normally N << M where M is the number of edges.

RickyBenevelli:
I agree with @matteo-dinoia . I don't think this solution works because it doesn't allow us to know all the arcs. Also having loops in the `path_trace` is not a problem. The path the packets should follow should not be decided on the single `path_trace` but on the graph reconstructed from all the `path_trace` of the `FloodingResponse`. 

Also note that any loops in the `path_trace` can be removed by removing only the last node in the `path_trace`.

PR #45, opened 2024-11-19
Refactoring protocol

matteo-dinoia:


Suge42 (approved):
These are just small refactoring changes in the protocol, I'd say they make sense

matteo-dinoia:
Ok will merge then

PR #51, opened 2024-11-20
Adding derive Clone and Debug to all struct

matteo-dinoia:
This cannot be done by the single groups, so we should do it for everyone so if someone need it for some struct it can use it.

### DIFF
--- a/crates/wg_config/src/config.rs
+++ b/crates/wg_config/src/config.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "serialize")]
 use serde::Deserialize;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


DiCesareDaniele:
I would also add PartialEq, Eq to ServerType and NodeType

matteo-dinoia:
> I would also add PartialEq, Eq to ServerType and NodeType

Fair, I'll add to any that make sense. For the rest being a stupidy small change do you thing, it is ok so we can merge it?

matteo-dinoia:
Other place we could add to is the message/packet, to some of the component but it seem excessive because in most case there is a unique id just for that.

DiCesareDaniele:
> Other place we could add to is the message/packet, to some of the component but it seem excessive because in most case there is a unique id just for that.

Nah that is overkill I think it's fine RN

PR #52, opened 2024-11-20
Small protocol refactoring

IlyaEmeliyanov:
Changes applied:
1. Removed the parts where the term "message" is used inappropriately or ambiguously and added clarity.
2. It’s logically better to treat the Query and QueryResult structs as a request-response pair. So they were renamed to QueryRequest and QueryResponse.
3. In fragment reassemble the assembling goes until `total_n_fragments` - 1 and not - 2. If it's still correct, can someone explain?
4. Some renaming and language adjustments.


AP-protocol.md
```diff
@@ -41,7 +41,7 @@ connected_drone_ids = ["connected_id1", "..."] # max 2 entries
 ```
 - note that `connected_drone_ids` cannot contain `client_id` nor repetitions
 - note that a client cannot connect to other clients or servers
-- note that a client can be connected to at most two drones
+- note that a client can be connected to at least one drone and at most two drones
```
matteo-dinoia, line 44: This may be better in another PR

AP-protocol.md
```diff
@@ -188,40 +188,40 @@ pub struct MessageData { // Only part fragmentized
...
+### Ack
+
+If a drone receives a Message and can forward it to the next hop, it also sends an Ack to the client.
+
+```rust
+pub struct Ack{
+	fragment_index: u64,
+	time_received: std::time::Instant
+}
+```
```
Suge42, line 200: I think that ACK are sent when the message is completed at the receiving end, not at every jump of every fragment.

AP-protocol.md
```diff
@@ -272,7 +272,7 @@ Note that, if there are more than one fragment, `file_size` must be 80 for all f
 
 If the client or server has already received a fragment with the same `session_id`, then it just needs to copy the data of the fragment in the vector.
 
-Once that the client or server has received all fragments (that is, `fragment_index` 0 to `total_n_fragments` -2), then it has reassembled the whole fragment.
+Once that the client or server has received all fragments (that is, `fragment_index` 0 to `total_n_fragments` - 1), then it has reassembled the whole fragment.
```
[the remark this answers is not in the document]
Suge42: Since is the faulty protocol, might be an error put originally by the professor on purpose.

AP-protocol.md
matteo-dinoia: That's fine but we should move. If we pile up PR it will be kinda of a mess to merge everything on monday.

IlyaEmeliyanov:
Can you submit the changes pls?

Suge42 (approved):

matteo-dinoia:
Sure but before change the name you changed also in the code.

IlyaEmeliyanov:
> Sure but before change the name you changed also in the code.

done

PR #53, opened 2024-11-20
Remove topology and impl message (which are group defined parts)

matteo-dinoia:
So move server type to message because it is still needed there


Marui-9:
Those present are just signatures and are operations that will be done on message / vec of frags. There is no reason to just throw them away. We will put them somewhere that makes sense, but for now we can leave them here. Can you just add server type instead of removing other code?

matteo-dinoia:
> of frags.

You are talking about impl Message.
1. they should be a trait
2. there is no reason to restrict the signature of a behavior that is client/server specific. If we want to restrict this we should just give a common implementation.

matteo-dinoia:
> Yes, it should be a trait, and we haven't defined it yet. That's why we can keep the signatures here, they can be modified and/or moved when we decide what to do about it.

I repeat: it is a part that the group will have define. We have no reason to define them, especially if we don't provide common code for that. 

At the moment is dead code so the best is just to remove it. So if you would like to propose a common code for this part, please open another PR with that.

DiCesareDaniele:
Love this. But I would completely remove wg_network since there is bearly anything there anymore and put ServerType, NodeType, NodeId, SourceRoutingHeader in Packet where I think they belong

DiCesareDaniele:
> Those present are just signatures and are operations that will be done on message / vec of frags. There is no reason to just throw them away. We will put them somewhere that makes sense, but for now we can leave them here. Can you just add server type instead of removing other code?

That is the Assembler part of the Protocol, which is not common code. 
Also I don't think it should be a trait since it's not a part that is shared

<3

Marui-9 (approved):

Suge42:
I think this is a good proposal, but since there's some doubt about it, is still probably better to quickly vote on it.

matteo-dinoia:
> ut since there's some doubt about it, is still probably better to quickly vote on it.

On WG or just on telegram? For smaller stuff like this i would prefers fixing it fast.

DiCesareDaniele:
> > ut since there's some doubt about it, is still probably better to quickly vote on it.
> 
> On WG or just on telegram? For smaller stuff like this i would prefers fixing it fast.

I would also suggest to vote on telegram since it's a small change, this way we can focus on more important stuff

Suge42:
Make sense, I'll make the vote tomorrow afternoon I think, so that without lectures people should see it quicker

PR #54, opened 2024-11-20
Add query and query result to packet types (+use NodeId)

matteo-dinoia:


DiCesareDaniele:
Nice. I think we can approve this without a vote since it was already part of the code inside protocol

DiCesareDaniele:
One thing I noticed is that this is not updating examples/drones also since this is a sync of what is present in the protocol and missing in the code I would remove FragmentData as @ZGako mentioned in #58

matteo-dinoia:
Everything should be fixed, please re-review.

DiCesareDaniele:
> Everything should be fixed, please @DiCesareDaniele (or anyone) re-review.

Yep, everything seems fine

PR #57, opened 2024-11-20
added packet_send channels in drone new() function

ZGako:
also changed drone_usage example to be more accurate


matteo-dinoia:
We were of the idea to just add the sender via the command of the simulation controller

DiCesareDaniele:
Worst case scenario one can leave the hashmap empty so I think it's a good change

PR #58, opened 2024-11-20
these changes were already approved 2 weeks ago

ZGako:
fixed the actual lib to support these changes


DiCesareDaniele:
> fixed the actual lib to support these changes

Good but it's a duplicate of #54

DiCesareDaniele:
> @DiCesareDaniele True but that one is missing the FragmentData removal

You are right but since they have the same purpose I would close this one and open a comment in #54 asking to remove FragmentData

matteo-dinoia:
Closing as we decided that this is duplicated of the other one and that the point in this PR where merged with the other

PR #60, opened 2024-11-21
Flood: remove TTL and SourceRoutingHeader

RickyBenevelli:
### Proposed Update to the Network Discovery Protocol

The current implementation of the Network Discovery Protocol is not well-standardized and contains redundancies. Below are the proposed updates:

1. **Removal of TTL**:  
   The `ttl` field is no longer necessary to determine the end of the flooding process. This is because drones already keep track of received `flood_id`s to avoid processing duplicate requests. This tracking mechanism naturally terminates the flooding process.  

2. **Flooding Termination Conditions**:  
   - A node does not have any neighbors to forward the `FloodRequest` to, other than the node from which it originally received the request.  
   - If this scenario occurs for a drone, it will not be utilized in the network (as it is effectively redundant). However, this situation can also occur for hosts that are connected to only one node.  

3. **Removal of the `SourceRoutingHeader` in `FloodResponse` Packets**:  
   The `SourceRoutingHeader` inside `FloodResponse` packets has been removed as it is redundant. The `Packet` structure already includes a `SourceRoutingHeader`, making the additional field unnecessary.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -119,9 +119,6 @@ struct FloodRequest {
 	flood_id: u64,
 	/// ID of client or server
 	initiator_id: NodeId,
-	/// Time To Live, decremented at each hop to limit the query's lifespan.
-	/// When ttl reaches 0, we start a FloodResponse message that reaches back to the initiator
-	ttl: u8,
 	/// Records the nodes that have been traversed (to track the connections).
 	path_trace: Vec<(NodeId, NodeType)>
 }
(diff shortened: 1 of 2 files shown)


matteo-dinoia:
You need to run cargo fmt --all

matteo-dinoia:
I like your changes but this is not a refactor, so i'll change the name

AP-protocol.md
```diff
@@ -130,28 +127,41 @@ struct FloodRequest {
...
-For every flood response or acknowledgment the initiator receives, it updates its understanding of the graph:
+For every flood response the initiator receives, it updates its understanding of the graph:
 
 - If the node receives a flood response with a **path trace**, it records the paths between nodes. The initiator learns not only the immediate neighbors but also the connections between nodes further out.
 - Over time, as the query continues to flood, the initiator accumulates more information and can eventually reconstruct the entire graph's topology.
 
 ### **Termination Condition**
 
 The flood can terminate when:
+- A node receives a `FloodRequest` with a flood_id that has already been received.
```
luiss07, line 163:
```suggestion
- A node receives a `FloodRequest` with a flood_id that has already been received.
```
If a drone builds only one `FloodResponse` from receiving multiple FloodRequest from different edges this algorithm does not allow to discover all the edges. Instead, if you create a `FloodResponse` every time that the `flood_id` has already been received but it has different incoming edge (not seen yet) we should be able to discover all of the available edges.

RickyBenevelli: yes, a `FloodResponse` is created each time an already received `flood_id` is received.

luiss07: If we say each time, without checking the incoming edge, then we might be sending redundant information to the flooding initiator. Anyway, i'd specify it, since I don't think this really clear this way.

PR #62, opened 2024-11-21
New message structure

daw-dev:
# New Message Structure
## `MessageContent` trait
As you can see, I created a trait `MessageContent`. This makes it so that you can treat different kinds of messages in different ways, so that for example `ChatServer` only need to know how to process `ChatRequest`s in order to create `ChatResponse`s.
## Example
Why should we accept this pull request?
Would you prefer something like
```rust
impl ChatServer {
    fn handle_request(&self, request: ChatRequest) {
        // simply handle chat-server requests
        unimplemented!()
    }

    fn send_response(&self, response: ChatResponse) {
        // simply handle chat-server responses
        unimplemented!()
    }
}
```
or something like
```rust
impl ChatServer {
    fn handle_request(&self, message_content: MessageContentSuperMegaEnumOrSmth) {
        match message_content {
            RqstChatClientList => unimplemented!(),
            RqstChatClientRegistration(id) => unimplemented!(),
            RqstTextFileList => panic!("WTF?"),
            RspnMedia(vec) => panic!("Why am I getting this?"),
            RqstToDoABackflip(wtf) => panic!("I don't know how to do a backflip 😭"),
            RspnWHAAAAAT => im_lost!("panic"),
        }
    }

    fn send_response(&self, message_content: MessageContentSuperMegaEnumOrSmth) {
        // same as above
        // don't make me find other stupid stuff...
        unimplemented!()
    }
}
```
(The first one, you would prefer the first one)
## Message check
f course somewhere the check that the request/response is of the right type should be somewhere, but where? Simple! The `MessageContent::deserialize` function should return a `Err(String)` (could be other type of errors).
## Server types
I have just a new issue with this system (otherwise basically perfect 🗿), how do we know what kind of client/server is a specific client/server? That's actually something to think about, either we include it in the flood or we create a request for every request/response type, such as `ChatRequest::IsChatServer` with `ChatResponse::IsChatServer(bool)`.
## Trait implications
Another thing to discuss is that, since this is a new trait, it makes it possible that every group just implements its own set of request/response.
## Work In Progress
This is just a draft, if this PR gets accepted, I will work on it more!

### DIFF
--- a/crates/wg_packet/src/message.rs
+++ b/crates/wg_packet/src/message.rs
@@ -1,69 +1,127 @@
-use crate::packet::Fragment;
-use wg_network::topology::ServerType;
 use wg_network::{NodeId, SourceRoutingHeader};
 
 #[derive(Debug, Clone)]
-pub struct Message {
-    pub message_data: MessageData,
+pub struct Message<M: MessageContent> {
+    pub message_data: MessageData<M>,
     pub routing_header: SourceRoutingHeader,
 }
 
 // Only part fragmentized
 #[derive(Debug, Clone)]
-pub struct MessageData {
+pub struct MessageData<M: MessageContent> {
     pub source_id: NodeId,
     pub session_id: u64,
-    pub content: MessageContent,
+    pub content: M,
+}
+
+pub trait MessageContent {
+    fn serialize(&self) -> String;
+    fn deserialize(serialized: String) -> Result<Self, String>
+    where
+        Self: Sized;
+}
+
+// ReqServerType,
+#[derive(Debug, Clone)]
+pub enum TextRequest {
+    FileList,
+    File(u64),
+}
+
+impl MessageContent for TextRequest {
+    fn serialize(&self) -> String {
+        match self {
+            TextRequest::FileList => "FileList".to_string(),
+            TextRequest::File(id) => format!("File({})", id),
+        }
+    }
+
+    fn deserialize(serialized: String) -> Result<Self, String> {
+        if serialized == "FileList" {
+            Ok(TextRequest::FileList)
+        } else if let Ok(id) = serialized
+            .trim_start_matches("File(")
+            .trim_end_matches(')')
+            .parse()
+        {
+            Ok(TextRequest::File(id))
+        } else {
+            Err("Failed to deserialize TextRequest".to_string())
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
-pub enum MessageContent {
-    // Client -> Server
-    ReqServerType,
-    ReqFilesList,
-    ReqFile(u64),
-    ReqMedia(u64),
-
-    ReqClientList,
-    ReqRegistrationToChat,
-    ReqMessageSend { to: NodeId, message: Vec<u8> },
-
-    // Server -> Client
-    RespServerType(ServerType),
-    RespFilesList(Vec<u64>),
-    RespFile(Vec<u8>),
-    RespMedia(Vec<u8>),
-    ErrUnsupporedRequestType,
-    ErrRequestedNotFound,
-
-    RespClientList(Vec<NodeId>),
-    RespMessageFrom { from: NodeId, message: Vec<u8> },
-    ErrWrongClientId,
+pub enum MediaRequest {
+    Media(u64),
 }
 
-impl Message {
-    // takes message and returns the data struct serialized in a String
-    // so it goes from the actual data struct to a String
-    #[allow(unused_variables)]
-    pub fn serialize(&self) -> String {
-        unimplemented!()
+impl MessageContent for MediaRequest {
+    fn serialize(&self) -> String {
+        todo!()
     }
-
-    // takes the content String and makes an instance of Message from it
-    #[allow(unused_variables)]
-    pub fn deserialize(serialized: String) -> Message {
-        unimplemented!()
+    fn deserialize(_: String) -> Result<Self, String> {
+        todo!()
     }
+}
 
-    // takes the String and splits it into Fragments
-    #[allow(unused_variables)]
-    pub fn disassembly(serialized: String) -> Vec<Fragment> {
-        unimplemented!()
+#[derive(Debug, Clone)]
+pub enum ChatRequest {
+    ClientList,
+    Register(NodeId),
+    SendMessage { to: NodeId, message: String },
+}
+
+impl MessageContent for ChatRequest {
+    fn serialize(&self) -> String {
+        todo!()
     }
+    fn deserialize(_: String) -> Result<Self, String> {
+        todo!()
+    }
+}
 
-    // takes a bunch of Fragments and composes them in a serialized string.
-    #[allow(unused_variables)]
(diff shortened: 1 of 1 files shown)


Marui-9:
Seems great! 👍

matteo-dinoia:
I don't see the appeal of this PR sorry. We can just use serde or something to serialize the messageContent and keep the original enum.

DiCesareDaniele:
RqstToDoABackflip is 🔥

daw-dev:
Well, @matteo-dinoia, the fact that you had to put `//Client->Server` kinda comments was always sus (usually when you specify something like this in comments it's because you could do things better), and the fact that Clients and Servers of X kind only know about X request and X responses is really good in my opinion. It's the really nice information hiding of `Good Code 101`.
With the big enum I could implement a `MediaServer` (as in `struct MediaServer`) that acts as a `ChatClient`, which is a really bad sign.

matteo-dinoia:
I don't see the appeal of this PR sorry. We can just use serde or something to serialize the messageContent and keep the original enum.

I'm mostly talking about your manual serialization

matteo-dinoia:
Also i think that doesn't simplify. Please provide an actual example of usage from Packet to usage of messageContent

luiss07:
Good work, we like it. But, we suggest to change `MessageContent` trait with a less misleading name, like `Serializable`.

daw-dev:
I wrote an example of how one could create a `Server` trait so that creating new servers becomes really easy. Also the cool thing is that if i create a new type of `TextRequest` (for example) with the big enum either all of my servers and clients or none would compile (depending if you implement everything by matching every single case or by matching with _), with my solution if I create a new `TextRequest` only my `TextServer` wouldn't compile and if I create a new `TextResponse` only my `TextClient` wouldn't compile (which I think is really good).

matteo-dinoia:
I repeat why are we reimplementing serialize when we could use serializable of Serde? Making it simpler and more in line with the other code?

crates/wg_packet/src/message.rs
```diff
@@ -1,69 +1,220 @@
...
+        todo!()
     }
+    fn deserialize(_: String) -> Result<Self, String> {
+        todo!()
+    }
+}
+
+impl Response for ChatResponse {}
+
+mod example {
```
matteo-dinoia, line 147: This should be in another file.

Marui-9:
How would that be more in line with other code? Implementing one function is not the end of the world. Plus it's only one data structure, we don't need an entire library.

matteo-dinoia:
The point is we already use (or at least allow the use of) serde for reading toml file.

Adding custom serialization, just add an extra breaking point for no apparent reason. Adding a lot of code when most of the people will be using serde anyway.

IlyaEmeliyanov:
I like this pr, especially because it makes the code look cleaner and modular. I have a couple of things to ask though:

1. What is `Text(u64)` and `Media(u64)`?
2. Can you clarify the meaning of the enum variants `MessageFrom` and `MessageSent`, so there are no doubts?
3. Also as @matteo-dinoia mentioned it's better to move the code example in the appropriate section.

And the last thing. This is more of a request for everyone who's submitting a pull request. 
If you're making changes to the code, please add some short comments to what you're implementing and why. It's not always trivial to understand what one's mind is thinking about. 
If you think that the naming that you're using is self-explanatory though, then it's ok imo to leave it without comments.
Let's follow some good practices on this.

Thanks <3

daw-dev:
I managed to put it in the examples folder

matteo-dinoia:
> Well, @IlyaEmeliyanov, actually this PR is really about the modularity of using different enums with common traits, so all the naming are just proof of concept. If this PR gets approved I'll try to make the enums in line with the ones that were there before.

Yeah, that is fine but do that before the merge.

> Also, while writing this code I noticed something... Should this even be in our common library? We share drones, and drones just like to talk about packets, acks, fragments, errors, drops and floods (they can be such gossips), so maybe messages should be a single group thing so that if @matteo-dinoia wants to create a big-ass enum he can do it (🫶🏻).

Joking aside I have no preference for the struct (splitted or not). But it probably should define them, because it could be useful in case of some drop out and because anyway we will need to do it, and it is not a very prescripting struct, if someone want to implement their custom serializable struct we could allow that.

Finally, I would like to ask again why can't serializable just extends Serde serialization or something. Why do we need to write the parsing by hand, creating more work for us and making it error prone, when there are already library for doing this?

daw-dev:
This PR is not about serialization, I don't know how does serde work, I'll look into it and I'll see if it fits well. In that case, you'll find another commit.

matteo-dinoia:
> This PR is not about serialization, I don't know how does serde work, I'll look into it and I'll see if it fits well. In that case, you'll find another commit.

Ok i understand. But a PR should be a single topics so it's easier and faster to vote and merge it. If something is not related to a PR, it shouldn't be in it. 

But no problem. Now, it's late so I guess we will vote without changing the PR (unless you try serde and find or find an even better solution). But we will split the vote I guess.

matteo-dinoia:
> > This PR is not about serialization, I don't know how does serde work, I'll look into it and I'll see if it fits well. In that case, you'll find another commit.
> 
> Ok i understand. But a PR should be a single topics so it's easier and faster to vote and merge it. This is expecially true the more open PR there are. If something is not related to a PR, it shouldn't be in it. 
> 
> But no problem. Now, it's late so I guess we will vote without changing the PR (unless you try serde and find or find an even better solution). But we will split the vote I guess.

crates/wg_packet/src/message.rs
```diff
@@ -1,69 +1,91 @@
-use crate::packet::Fragment;
-use wg_network::topology::ServerType;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
 use wg_network::{NodeId, SourceRoutingHeader};
 
 #[derive(Debug, Clone)]
-pub struct Message {
-    pub message_data: MessageData,
+pub struct Message<M: DroneSend> {
```
luiss07, line 6: I think that DroneSend is confusing. Because we are using it also to deserialize and not only to "send". I'd prefer `Serializable` as before, just because it means that it contains both serialize and deserialize (even if now they are `stringify` and `from_string`). Or if you can come up with something else more generic it should be fine anyway.

daw-dev: I was exploring different namings, I was trying to make it so that that trait just implies "this can be sent with drones".

daw-dev:
Another important change is that now message doesn't include the source routing header, this is to separate high level and low level and also to make it so that different fragment of the same message can be routed differently (which I think is cool).

matteo-dinoia:
Thank you for the commit, this is more readable and easier. Obviously we will need to vote but I think now it's ok.

DiCesareDaniele:
> I think this is merge-able, tell me if I have to make some last small adjustments.

Pls, move serde and serd_json from dependencies to dev-dependencies. Thanks <3

matteo-dinoia:
@DiCesareDaniele We decided to remove the entire file.

matteo-dinoia:
@Suge42 can i merge this?

PR #66, opened 2024-11-21
Refine Source Routing Protocol

RickyBenevelli:
The updates aim to enhance clarity, correctness, and robustness of the protocol by:
1. Clarifying Protocol Steps: Providing a detailed and accurate description of each step in the protocol, ensuring consistent usage of `hop_index` and eliminating potential off-by-one errors.
2. Standardization of routing-related error handling.
3. Improving the example to make it more understandable and easily implemented

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -84,6 +84,28 @@ When a client or server wants to send a message to another node, it performs the
 
 3. **Packet Sending**: The sender attaches the source routing header to the packet and sends it to the first node in the route (the node at `hops[1]`).
 
+
+### Source Routing Protocol Steps
+
+1. Initialization:
+	- Sender sets `hop_index` to **1**.
+	- Constructs `hops` list including the sender and all intermediate nodes to the destination.
+
+2. At Each Node:
+	- **Step 1**: Check if `hops[hop_index]` matches the node's own ID.
+		- **If yes**, proceed to Step 2.
+		- **If no**, send a Nack with `ErrorInRouting` and terminate processing.
+	- **Step 2**: Increment `hop_index` by **1**.
+	- **Step 3**: Determine if the node is the final destination:
+		- **If `hop_index` equals the length of `hops`**, the node is the final destination and processes the packet.
+		- **If not**, proceed to Step 4.
+	- **Step 4**: Identify the next hop using `hops[hop_index]`.
+		- **If the next hop is not a neighbor**, send a Nack with `ErrorInRouting` indicating the problematic node ID.
+		- **If the next hop is a neighbor**, send the packet to the next hop.
+
+3. Final Destination:
+	- The node processes the packet as it has reached its intended recipient.
+
 ### Step-by-Step Example
 
 Consider the following simplified network topology:
@@ -99,17 +121,6 @@ Suppose that client A wants to send a message to server D.
 	- **`hops`**: `[A, B, E, F, D]`.
 	- **`hop_index`**: `1`.
 - Sends the packet to **B**, the first node after itself.
-
-**At Each Node**:
-
-- **Receiving the Packet**:
-	- The node receives the packet and checks if it is the intended recipient by comparing its own ID with `hops[hop_index]`.
-		- **If they match**, the node proceeds.
-		- **If they do not match**, the node identifies a routing error and sends back a **Nack** (Negative Acknowledgment) with an `ErrorInRouting` indicating the node ID where the error occurred.
-- **Processing the Packet**:
-	- Increments `hop_index` by **1** to move to the next hop.
-	- Determines the next hop using the updated `hop_index`.
-	- Sends the packet to the next hop.
 
 **Detailed Steps**:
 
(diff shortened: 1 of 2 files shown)


crates/wg_network/src/routing.rs
```diff
@@ -2,7 +2,7 @@ pub type NodeId = u8;
 
 #[derive(Debug, Clone)]
 pub struct SourceRoutingHeader {
-    pub hop_index: usize, // must be set to 0 initially by the sender
+    pub hop_index: usize, // must be set to 1 initially by the sender
```
matteo-dinoia, line 5: This is not true because of the drone rule

RickyBenevelli: I did not understand what you are referring to. Why shouldn't that be true because of the drone rules?

IlyaEmeliyanov (approved):
looks good to me

crates/wg_network/src/routing.rs
matteo-dinoia: Because the drone increase by one the index as first rules.

matteo-dinoia: No because the array is 0 indexed, so if we take your patch, on the first hop the index would be increase to 2 and then the next we use to decide to do forwarding is 3 (which is wrong).

If you also change the rules that will be fine, but maybe that could be better in another PR.

AP-protocol.md
```diff
@@ -278,29 +315,86 @@ Once that the client or server has received all fragments (that is, `fragment_in
...
-2. obtain the (new `hop_index`) + 1 element of the `SourceRoutingHeader` vector `hops`, let's call it `next_hop`
-	* It **must ignore** intentionally to check `hop_index`.
+When a drone receives a packet, it **must** perform the following steps:
 
-3. if `next_hop`
-	* doesn't exist create a new packet of type Nack, precisely of type `DestinationIsDrone`. The packet must have the routing made of a vector but inverted and only contains the nodes from this drone to the sender. Send this packet as a normal packet. End here.
-	* if the `NodeId` is not a neighbor, then creates a new packet of type Nack, precisely of type `ErrorInRouting` with field the value of `NodeId` of next hop. Continue as other error.
+1. **Step 1**: Check if `hops[hop_index]` matches the drone's own `NodeId`.
+	- **If yes**, proceed to Step 2.
+	- **If no**, send a Nack with `ErrorInRouting` (including the drone's own `NodeId`) and terminate processing.
```
matteo-dinoia, line 323: I would create another error types because if the drone send this error the client may think it that the drone were removed.

AP-protocol.md
```diff
@@ -278,29 +315,86 @@ Once that the client or server has received all fragments (that is, `fragment_in
...
 
-4. Proceed as follows based on packet type:
+2. **Step 2**: Increment `hop_index` by **1**.
 
-### Flood Messages
-TODO  (If the packet is flood related, follow the rules in the flood section)
+3. **Step 3**: Determine if the drone is the final destination:
+	- **If `hop_index` equals the length of `hops`**, the drone is the final destination and processes the packet accordingly.
```
matteo-dinoia, line 328: Processing here means always sending an error. Because this is not specified anywhere else in the protol please add it back here.

matteo-dinoia (commented):
Seem good, just a few notes to change.

AP-protocol.md
```diff
@@ -66,33 +66,70 @@ Recall that there are: Content servers (that is, Text and Media servers) and Com
...
+
+When a client or server wants to send a message to another node, it performs the following steps:
+
+1. **Route Computation**: The sender calculates the entire path to the destination node. This path includes the sender itself and all intermediate nodes leading to the destination.
+
+2. **Creation of the Source Routing Header**: The sender constructs a header that contains:
+	- **`hops`**: A list of node IDs representing the route from the sender to the destination.
+	- **`hop_index`**: An index indicating the current position in the `hops` list. It starts at **1** because the first hop (`hops[0]`) is the sender itself.
+
+3. **Packet Sending**: The sender attaches the source routing header to the packet and sends it to the first node in the route (the node at `hops[1]`).
```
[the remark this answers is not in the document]
RickyBenevelli: I don't see any other possible order

AP-protocol.md
RickyBenevelli: yes, you are right

crates/wg_network/src/routing.rs
RickyBenevelli: I’m closing the conversation because I believe it was based on a partial reading of the changes. Feel free to reopen it if I misunderstood something.

AP-protocol.md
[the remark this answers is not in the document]
matteo-dinoia: Statistically we already have the sourceRoutingHeader created expecially if the node already tried to send the packet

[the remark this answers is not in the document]
RickyBenevelli: yes, that's right. In this case, steps 1 and 2 would already be done

matteo-dinoia:
Also, I have seen you added quite the bit. I think to not have this document became infinite, it may be best to move the examples to their own file (maybe just a `example.md` which contains all non code examples)

AP-protocol.md
```diff
@@ -66,33 +66,70 @@ Recall that there are: Content servers (that is, Text and Media servers) and Com
...
+	- **`hop_index`**: `1`.
+- Sends the packet to **B**, the first node after itself.
+
+**At Each Node**:
 
-It computes the route B→E→F→D, creates a **Source Routing Header** specifying route A→B→E→F→D, adds it to the packet and sends it to B.
+1. **Node B**:
+	- Receives the packet.
+	- Determines that the next hop is **E**.
+	- Sends the packet to **E**.
```
luiss07, line 110:
```suggestion
1. **Node B**:
	- Receives the packet.
	- Increments the **`hop_index`** by 1.
	- Determines that the next hop is **E**.
	- Sends the packet to **E**.
```
This needs to be applied also to the following steps.

RickyBenevelli: yes, it was to not shift the focus to the Drone Protocol but to leave the focus of the example on the Source Routing Protocol. I will now however move everything to an example file as suggested by @matteo-dinoia

DiCesareDaniele:
> Also, I have seen you added quite the bit. I think to not have this document became infinite, it may be best to move the examples to their own file (maybe just a `example.md` which contains all non code examples)

Maybe in a separate PR. So we can move all the examples

AP-protocol.md
RickyBenevelli: @DiCesareDaniele is right, better to move the examples in a separate PR

PR #67, opened 2024-11-22
Check for bidirectionality when initializing network

marcoff181:
Formalizes the steps that the network initializer takes, especially checking that the network initialization file represents a valid network topology

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -14,9 +14,15 @@ type NodeId = u64;
 
 # Network Initializer
 
-The **Network Initializer** reads a local **Network Initialization File** that encodes the network topology and the drone parameters and, accordingly, spawns the node threads and sets up the Rust channels for communicating between nodes.
-
-> Importantly, the Network Initializer should also setup the Rust channels between the nodes and the Simulation Controller (see the Simulation Controller section).
+The **Network Initializer**:
+1. reads a local **Network Initialization File** that encodes the network topology and the drone parameters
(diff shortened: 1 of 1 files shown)


AP-protocol.md
```diff
@@ -14,9 +14,15 @@ type NodeId = u64;
...
-> Importantly, the Network Initializer should also setup the Rust channels between the nodes and the Simulation Controller (see the Simulation Controller section).
+The **Network Initializer**:
+1. reads a local **Network Initialization File** that encodes the network topology and the drone parameters
+2. checks that the initialization file adheres to the formatting and restrictions defined in the section below 
+3. checks that the initialization file represents a bidirectional graph
+4. according to the network topology defined in the initialization file:
+   - spawns the node threads
+   - spawns the simulation controller thread
+   - sets up the Rust channels for communicating between nodes that are connected in the topology
+   - sets up the Rust channels for communication between nodes and the simulation controller
```
matteo-dinoia, line 25: Only problem is order. Spawning nodes thread should probably be almost last. 
I would recommend writing somewhere that these substep don't need to be done in order.

matteo-dinoia (commented):
For the rest is good.

DiCesareDaniele:
> That seem totally fine to me, but not sure if we should vote on telegram. @Suge42 what do you think?

I would not vote since it does not change the code. It just adds documentation

Suge42:
We'll vote this during the next WGM, mainly to notify everyone about it and answer eventual doubts

PR #68, opened 2024-11-22
Removed ambiguity in Fragment Reassembly

### LINKED ISSUE #59: Ambiguous terms in Fragment Reassembly
In the Fragment Reassembly paragraph there is the terms `file_size` and `file` and it is the only place where I found them.
I think they are namely `length: u8` and `data: [u8; 80]` attributes of the `struct Fragment` but I am not sure.

Below there is the cited paragraph.

> ```rust
> // OTHER CODE
> pub struct Fragment {
>	 fragment_index: u64,
>	 total_n_fragments: u64,
>	 length: u8,
>	 // assembler will fragment/defragment data into bytes.
> 	data: [u8; 80] // usable for image with .into_bytes()
> }
> ```
>
>  To reassemble fragments into a single packet, a client or server uses the fragment header as follows:
> 1. The client or server receives a fragment.
> 2. It first checks the `session_id` in the header.
> 3. If it has not received a fragment with the same `session_id`, then it creates a vector big enough where to copy the data of the fragments.
> 4. The client would need to create a `vec<u8>` with capacity of `total_n_fragments` * 80.
> 5.  It would then copy `file_size` elements of the `file` array at the correct offset in the vector.
> Note that, if there are more than one fragment, `file_size` must be 80 for all fragments except for the last.

Isax03:
The PR seeks to resolve the ambiguity introduced by the terms *`file` array* and *`file_size`* within the context of the Fragment Reassembly paragraph.
Additionally, points 3 and 4 have been consolidated, as they both pertain to the same step in the reassembly process.

Fixes #59

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -263,13 +263,13 @@ To reassemble fragments into a single packet, a client or server uses the fragme
 
 2. It first checks the `session_id` in the header.
 
-3. If it has not received a fragment with the same `session_id`, then it creates a vector big enough where to copy the data of the fragments.
+3. If it has not received a fragment with the same `session_id`, then it creates a vector (`Vec<u8>` with capacity of
+   `total_n_fragments` * 80) where to copy the
+   data of the fragments;
 
-4. The client would need to create a `vec<u8>` with capacity of `total_n_fragments` * 80.
+4. It would then copy `length` elements of the `data` array at the correct offset in the vector.
 
-5.  It would then copy `file_size` elements of the `file` array at the correct offset in the vector.
-
-Note that, if there are more than one fragment, `file_size` must be 80 for all fragments except for the last.
+> Note: if there are more than one fragment, `length` must be 80 for all fragments except for the last.
 
 If the client or server has already received a fragment with the same `session_id`, then it just needs to copy the data of the fragment in the vector.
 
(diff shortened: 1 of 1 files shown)


AP-protocol.md
```diff
@@ -354,12 +354,12 @@ pub enum MessageContent {
...
 
 	// Server -> Client
-	RespServerType(ServerType)
+	RespServerType(ServerType),
 	RespFilesList(Vec<u64>),
 	RespFile(Vec<u8>),
 	RespMedia(Vec<u8>),
 	ErrUnsupporedRequestType,
-	ErrRequestedNotFound
+	ErrRequestedNotFound,
```
matteo-dinoia, line 362: Please revert this to avoid conflict, someone else is already working on this in other PR.

matteo-dinoia (approved):
It is fine, just a minor git related issue, detailed below.

AP-protocol.md
```diff
@@ -263,13 +263,13 @@ To reassemble fragments into a single packet, a client or server uses the fragme
 
 2. It first checks the `session_id` in the header.
 
-3. If it has not received a fragment with the same `session_id`, then it creates a vector big enough where to copy the data of the fragments.
+3. If it has not received a fragment with the same `session_id`, then it creates a vector (`Vec<u8>` with capacity of
+   `total_n_fragments` * 80) where to copy the
+   data of the fragments;
 
-4. The client would need to create a `vec<u8>` with capacity of `total_n_fragments` * 80.
+4. It would then copy `length` elements of the `data` array at the correct offset in the vector.
```
matteo-dinoia, line 270: the `length` elements of the `data` array is a bit confusing.

Maybe something like copy the N elements of the `data` array where N is the `length`. Or something, but I think it could be clearer.

Fix this and I'll merge it, because there is nothing to vote on (just using right names

matteo-dinoia: Ok if I'm the only one which find it not super clear at first read than is fine like this

Suge42:
I'd say this can be merged @matteo-dinoia

PR #69, opened 2024-11-22
Update min number drones for client

IlyaEmeliyanov:
As suggested, this small change was moved to a separate pr.
The protocol doesn't explicitly specify the min number of drones the client should be connected to, so I'm suggesting to set it to 1 and call it a day.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -41,7 +41,7 @@ connected_drone_ids = ["connected_id1", "..."] # max 2 entries
 ```
 - note that `connected_drone_ids` cannot contain `client_id` nor repetitions
 - note that a client cannot connect to other clients or servers
-- note that a client can be connected to at most two drones
+- note that a client can be connected to at least one and at most two drones
 
 ### Servers
 Any number of servers, each formatted as:


matteo-dinoia:
I don't think is a priority so we can vote on next lesson.

DiCesareDaniele:
> I don't think is a priority so we can vote on next lesson.

The main document says that it's always possible for a client to reach a server so I think it's not adding any assumptions that need to be voted on

> we assume that network partitions that
separate clients from servers never happen so that it is always possible to
eventually deliver packets

IlyaEmeliyanov:
> > I don't think is a priority so we can vote on next lesson.
> 
> The main document says that it's always possible for a client to reach a server so I think it's not adding any assumptions that need to be voted on
> 
> > we assume that network partitions that
> > separate clients from servers never happen so that it is always possible to
> > eventually deliver packets

I know but adding clarity is never bad.

DiCesareDaniele:
> I know but adding clarity is never bad.

NoNo I like the PR just saing that we do not need to vote on this I would merge immediately. We have 10 PR open the more we can merge the easier it is to keep track of the others

DiCesareDaniele:
> We have 10 PR open the more we can merge the easier it is to keep track of the others

+2 issues +2 PR I am going to open later 😂

matteo-dinoia:
Merging then

DiCesareDaniele:
> Merging then

Nice

PR #70, opened 2024-11-22
removed time::Instant

DiCesareDaniele:
## Remove time::Instant
Remove time::Instant from Ack and Nack. Why are they even there, they are not realistic in a real world scenario and they are not tipical values that you can send through a network. Plus they do not serve any use, change my mind.


matteo-dinoia:
Fair

matteo-dinoia:
Approved by vote on TG

PR #71, opened 2024-11-22
Drone ids

DiCesareDaniele:
## Config
Ambiguous name connected_drone_ids in wg_2024::config::Drone but a drone can be connected to clients and server.
I renamed connected_drone_ids to connected_node_ids

Also if you did
```bash
cargo run --example parser --features="serialize"
```

before it would panic cause the path of config.toml was wrong. Got fixed


IlyaEmeliyanov:
Reasonable change, I think we can merge this.

matteo-dinoia:
Yeah, merging it

PR #72, opened 2024-11-22
Simulation Controller commands specification

luiss07:
The purpose of this PR is to define the commands that can be exchanged between the nodes (clients, servers and drones) and the Simulation Controller. 

The commands that I identified are the following: 
```rust
pub enum Command {
    AddChannel(NodeId, Sender<Packet>),
    RemoveChannel(NodeId),
    Crash,
    ChangePdr(f32),
    SentMessage(Packet),
    DroppedMessage(Packet)
}
```

You can check the command explanation directly from the PR, thus I won't put it here. My idea is to keep only one `Command` enum to avoid the definition of multiple channels. 

In the specification I clearly stated in which direction each message can be sent, therefore if the WG prefers to split the Commands it would be a small job.

**P.S.**: for the `Crash` command I wrote the following: `This command makes a node crash. Upon receiving this command, the node’s thread should return as soon as possible. The last thing the node should do is to confirm that it is terminating by sending a Crash command to the SC. ` The last part is just an idea of how we could get the confirmation that the node is terminating instead of periodically checking from the SC that the node crashed. I'm open to other ideas.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -307,30 +307,42 @@ TODO
 
 # Simulation Controller
 
-Like nodes, the **Simulation Controller** runs on a thread. It must retain a means of communication with all nodes of the network, even when drones go down.
+Like nodes, the **Simulation Controller** (SC) runs on a thread. It must retain a means of communication with all nodes of the network, even when drones go down. 
+The Simulation controller can send different commands to the nodes (drones, clients and servers) through a reserved channel. The **command** list of available commands is as follows:
+
+```rust
+pub enum Command {
+    AddChannel(NodeId, Sender<Packet>),
+    RemoveChannel(NodeId),
+    Crash,
+    ChangePdr(f32),
+    SentMessage(Packet),
+    DroppedMessage(Packet)
+}
+```
 
 ### Simulation commands
 
-The Simulation Controller can send the following commands to drones:
+Below are documented the commands that can be exchanged between the SC and any type of node (drone, client, server):
 
-`Crash`: This commands makes a drone crash. Upon receiving this command, the drone’s thread should return as soon as possible.
+- **SC -> Node (any)**
 
-`AddSender(crossbeam::Sender, dst_id)`: This command provides a node with a new crossbeam Sender to send messages to node `dst_id`.
+  - `Crash`: This commands makes a node crash. Upon receiving this command, the node’s thread should return as soon as possible. The last thing the node should do is to confirm that it is terminating by sending a Crash command to the SC. (Possible idea, other proposal are accepted). 
+  - `AddChannel(NodeId, Sender<Packet>)`: This command tells a node to add a new **edge** to the node with id `NodeId` and to use the provided `Sender<Packet>` channel to communicate with it. Since we are assuming bidirectional communication, this command must be used **two times** (i.e. to add the edge from `A` to `B`, we must add the `Sender from A to B` and the `Sender from B to A`).
+  - `RemoveChannel(NodeId)`: This command tells a node to remove the edge to the node with id `NodeId`. Since we are assuming bidirectional communication, this command must be used **two times** (i.e. to remove the edge from `A` to `B`, we must remove the `Sender from A to B` and the `Sender from B to A`).
 
-`AddReceiver(mpsc::Receiver, src_id)`: This command provides a node with a new crossbeam Receiver to receive messages from node `src_id`.
+- **Node (any) -> SC**
 
-`Spawn(id, code)`: This command adds a new drone to the network.
+  - `Crash`: used by a node to confirm that it is terminating.
+  - `SentMessage(Packet)`: a node sends this command every time it sends a message to another node (drone, client or server). In this way, the SC can keep track of the network activity.
 
-`SetPacketDropRate(id, new_pdr)`:
+Additionally, to the commands above, the SC and the drones can **exclusively exchange** the following commands:
 
-### Simulation events
+- **SC -> Drone**
+  - `ChangePdr(f32)` this command tells a drone to change its Packet Drop Rate to the new value provided.
 
-The Simulation Controller can receive the following events from nodes:
-
-`Topology(node_id, list_of_connected_ids, metadata)`: This event indicates that node `node_id` has been added to the network and its current neighbors are `list_of_connected_ids`. It can carry metadata that could be useful to display, such as the PDR and DR of Drones.
-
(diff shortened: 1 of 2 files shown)


matteo-dinoia:
Please run rustfmt --all

AP-protocol.md
```diff
@@ -307,30 +307,42 @@ TODO
...
-`Crash`: This commands makes a drone crash. Upon receiving this command, the drone’s thread should return as soon as possible.
+```rust
+pub enum Command {
+    AddChannel(NodeId, Sender<Packet>),
+    RemoveChannel(NodeId),
+    Crash,
+    ChangePdr(f32),
+    SentMessage(Packet),
+    DroppedMessage(Packet)
+}
```
matteo-dinoia, line 321: I would split in two enum

matteo-dinoia (commented):
Fine just some minor issue

AP-protocol.md
Suge42: I don't think we need to split the enum for now, maybe later in the implementation, but since it needs to be a unique type to be sent in the same channel, for now this doesn't add too much complexity.

AP-protocol.md
```diff
@@ -307,30 +307,42 @@ TODO
...
-
-`Crash`: This commands makes a drone crash. Upon receiving this command, the drone’s thread should return as soon as possible.
+```rust
+pub enum Command {
+    AddChannel(NodeId, Sender<Packet>),
+    RemoveChannel(NodeId),
+    Crash,
+    ChangePdr(f32),
+    SentMessage(Packet),
+    DroppedMessage(Packet)
```
Suge42, line 320: I would names into 'SentPacket' and 'DroppedPacket' or fragment, since the message should be the whole thing, and the drones only transports the fragments.
I'd also add the id of the drone the packet is sent to inside 'SentMessage', so that the sim controller can easily display the sending and receiving drone with the message.
And the pdr in 'DroppedMessage' might be useful to display too.

AP-protocol.md
matteo-dinoia: The point is that if we split in two we don't have to check extra illegal cases.

Becuase we would have a channel of commandFromSC e un channel of commandFromNode (names are very indicatives)

AP-protocol.md
luiss07: I agree with you, Ill change them and add the PDR in DroppedMessage

AP-protocol.md
luiss07: You would just need to add a _ in the match case. I get your point but I don't really see the trade-off with having even more channels to manage.

AP-protocol.md
```diff
@@ -307,30 +307,42 @@ TODO
...
 
-`AddReceiver(mpsc::Receiver, src_id)`: This command provides a node with a new crossbeam Receiver to receive messages from node `src_id`.
+Below are documented the commands that can be exchanged between the **SC and any type of node** (drone, client, server):
 
-`Spawn(id, code)`: This command adds a new drone to the network.
+- **SC -> Node (any)**
 
-`SetPacketDropRate(id, new_pdr)`:
+  - `Crash`: This commands makes a node crash. Upon receiving this command, the node’s thread should return as soon as possible. The last thing the node should do is to confirm that it is terminating by sending a Crash command to the SC. 
+  - `AddChannel(NodeId, Sender<Packet>)`: This command tells a node to add a new **edge** to the node with id `NodeId` and to use the provided `Sender<Packet>` channel to communicate with it. Since we are assuming bidirectional communication, this command must be used **two times** (i.e. to add the edge from `A` to `B`, we must add the `Sender from A to B` and the `Sender from B to A`).
```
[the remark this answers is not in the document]
luiss07: That PR is about adding a way to store the different Sender channels in a drone. When a drone receives the AddChannel command, it will add that Sender to the HashMap. Or is the drone initialized with all the available Senders (I hope it is not intended this way)?

AP-protocol.md
```diff
@@ -307,30 +307,42 @@ TODO
...
+  - `RemoveChannel(NodeId)`: This command tells a node to remove the edge to the node with id `NodeId`. Since we are assuming bidirectional communication, this command must be used **two times** (i.e. to remove the edge from `A` to `B`, we must remove the `Sender from A to B` and the `Sender from B to A`).
```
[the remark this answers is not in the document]
luiss07: If a drone has more edges, why would you crash the other end? Then, all the nodes connected to the crashed drone would have dangling senders.

AP-protocol.md
luiss07: I just noticed that within the SourceRoutingHeader (in Packet) we have the hop_index, thus we can retrieve the Node involved in SentPacket and DroppedPacket from there. 

For now, ill just add the PDR to DroppedPacket and refactor their names. Then let me know what you think.

AP-protocol.md
```diff
@@ -307,30 +307,42 @@ TODO
...
 
-`Topology(node_id, list_of_connected_ids, metadata)`: This event indicates that node `node_id` has been added to the network and its current neighbors are `list_of_connected_ids`. It can carry metadata that could be useful to display, such as the PDR and DR of Drones.
+In addition to the commands above, the **SC** and the **drones** can **exclusively exchange** the following commands:
 
-`MessageSent(node_src, node_trg, metadata)`: This event indicates that node `node_src` has sent a message to `node_trg`. It can carry useful metadata that could be useful display, such as the kind of message, that would allow debugging what is going on in the network.
+- **SC -> Drone**
+  - `ChangePdr(f32)` this command tells a drone to change its Packet Drop Rate to the new value provided.
 
+- **Drone -> SC**
+  - `DroppedPacket(Packet, f32)`: the drone sends this command every time it drops a packet. The `f32` value is the *Packet Drop Rate* of the drone at the time the packet was dropped.
```
matteo-dinoia, line 345: Not sure we need a PDR in this command

luiss07: I added it because yesterday I got suggested that it might be useful to display on the SC when a packet is dropped. Since it does not add any complication, I think it is fine to have it for those who want to use it

matteo-dinoia: Yeah but the SC knos the pdr already

luiss07: Right, I guess it depends on the group implementation but somewhere in their topology they should have the PDR (which at this point I think should be changed accordingly when using the Chandepdr command). Ill revert it back and close this review. 

PS: If you can check my other responses it would be fine, so I can review / close also those.

AP-protocol.md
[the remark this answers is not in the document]
luiss07: Im closing these two conversations, leaving open only the one about splitting the enum.

matteo-dinoia:
Vote for other PR

PR #73, opened 2024-11-22
Cargo should auto-update wg_repo dependencies

Sm3r:
Added brach param

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,6 @@ Please read [CONTRIBUTE.md](CONTRIBUTE.md)
 _Cargo.toml_
 ```toml
 [dependencies]
-wg_2024 = { git = "https://github.com/WGL-2024/WGL_repo_2024.git", features = ["serialize"] }
+wg_2024 = { git = "https://github.com/WGL-2024/WGL_repo_2024.git", branch = "main", features = ["serialize"] }
 ```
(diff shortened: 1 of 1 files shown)


DiCesareDaniele:
Cargo already uses main as the default branch

matteo-dinoia:
The problem is about the fact that the dependencies like this don't auto updates. We should add a command to update the `wg_repo` dependency.

DiCesareDaniele:
Yeah but adding main as branch does not auto-update

See
https://github.com/rust-lang/cargo/pull/8561
> Once a `git` dependency has been added, Cargo will lock that dependency to the
latest commit at the time. New commits will not be pulled down automatically
once the lock is in place. However, they can be pulled down manually with
`cargo update`.

And
https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/specifying-dependencies.md?plain=1#L274C1-L275C65

> Cargo locks the commits of git dependencies in Cargo.lock file at the time of their addition and checks for updates only when you run cargo update command

matteo-dinoia:
Sure yes, just saying we should add this to the README.md (the thing you are saying).

DiCesareDaniele:
> Sure yes, just saying we should add this to the README.md (the thing you are saying).

Sorry my bad

PR #76, opened 2024-11-23
Revise the flooding protocol, also merging the relative nodetypes etc…

franko-fox:
… in the flood.rs file and adding a termination condition.

This overhaul removes the previous TTL system, and improves message passing by only requiring a total of V+E messages to be passed, where V is the number of drones and E is the number of connections.

The algorithm pretty much works by building a spanning tree within the network, and removes the path tracing and source routing headers, opting instead for simpler HashSets which can be merged (the Add trait is already implemented to do most of the job of merging replies), at the cost of requiring the drones to store some more information.

An option for broadcasting the addition of a new drone is also present, as is a "get path to X" feature which can be useful, for ex. in case of a crash to quickly find a valid connection in less time than a complete network flood.

Some options in the flooding protocol are optional and can be left unimplemented (for example, the "to" field can simply be ignored without causing any significant difference in returned replies, and drones don't have to broadcast themselves at init). There are also some performance considerations which can be chosen to be applied or not by groups.


crates/wg_network/src/routing.rs
```diff
@@ -1,4 +1,4 @@
-pub type NodeId = u8;
+pub type NodeId = u64;
```
matteo-dinoia, line 1: We voted rhat u8 was fine

crates/wg_network/src/node.rs
```diff
@@ -5,11 +5,6 @@ use std::rc::Rc;
 
 pub type NodeRef = Rc<RefCell<Node>>;
 
-#[derive(Debug, Clone)]
-pub struct Topology {
-    pub nodes: Vec<NodeRef>,
-}
-
```
matteo-dinoia, line 12: There is another PR that is removing this part. Please stick only to one argument per PR.
Same with the other file you deleted.

So please revert these changes.

matteo-dinoia (commented):
Please start by removing the changes that are not related to tbis PR. I'll read the rest when i can.

RickyBenevelli:
I personally really like some of the ideas you had, however I also see some critical issues. I'll leave the comments below

RickyBenevelli:
I really like some of your changes, however there are some possible errors in my opinion:

What happens if one of the neighbors is not responding with FloodProcessing but is also not crashed?
In the example below, node `A` receives from “...” the `FloodRequest` (and responds with a `FloodProcessing`), sends the `FloodRequest` to `C` (which responds to it with a `FloodProcessing`). Before `A` can send the `FloodRequest` to `B` as well, `B` receives it from `C`, and then responds to `C` with a `FloodProcessing` and ignores the `FloodRequest` from `A`. At this point, however, `A` knows that not all of its neighbors are done or failed, because `B` is neither, so `A` will block the flooding. 

image

AP-protocol.md
```diff
@@ -106,54 +106,73 @@ When the network is first initialized, nodes only know who their own neighbors a
...
+### **Special considerations for Requests containing a "to" field**
+In this case, when returning a `FloodData` packet, the fields `connections` and `node_list` get set to empty, unless:
+	- The node is the target node, in which case it MUST populate `node_list` with itself, and MUST add the connection to the previous hop on the request to `connections`, or
+	- The node is not the target node but has received a non-empty `FloodData`, in which case it MUST add the connection to the previous hop as described above, and CAN add itself to the `node_list`.
 
-### **Termination Condition**
+### **Broadcast messages**
 
-The flood can terminate when:
+When a drone first connects to the network, it creates a `FloodData` message, containing a random `flood_id`, itself as the `node_list` and its neighbors as `connections`, with `is_broadcast` set to true. This message SHOULD get forwarded to neighbouring nodes. When receiving such a message a node MUST forward it to every other neighbor except the one it received it from. If received a second time the message MUST get ignored.
```
RickyBenevelli, line 171: I like this, it greatly simplifies flooding management, however in theory drones should not be able to initiate flooding. Please correct me if I am wrong

AP-protocol.md
```diff
@@ -106,54 +106,73 @@ When the network is first initialized, nodes only know who their own neighbors a
...
+The `connections` field gets populated with the neighbors, creating a tuple of two node IDs, one being the current and one being the neighbor, which MUST be in ascending order: (n1,n2) where n1 < n2 always.
+The node then proceeds to merge all obtained responses, and MUST send the final response to the previous hop it had stored, changing the `from` field to its own ID.
 
-### **Recording Topology Information**
-
-For every flood response or acknowledgment the initiator receives, it updates its understanding of the graph:
-
-- If the node receives a flood response with a **path trace**, it records the paths between nodes. The initiator learns not only the immediate neighbors but also the connections between nodes further out.
-- Over time, as the query continues to flood, the initiator accumulates more information and can eventually reconstruct the entire graph's topology.
+### **Special considerations for Requests containing a "to" field**
```
RickyBenevelli, line 164: I do not believe that this solution is more efficient than the general solution. In fact, the whole network will still have to be explored to wait for all the neighbors to be done or failed

RickyBenevelli: Perhaps one could change the algorithm in case the `to` field is set: a drone could return the `FloodResponse` as soon as it gets a non-empty one, without waiting for all `FloodProcessing` to be resolved.

AP-protocol.md
franko-fox: That can be discussed at the next WG I guess, it would make it easier for clients and servers to respond to new drone additions without requiring another flood of the whole network. (This part of the proposal isn't mandatory, so it can be skipped if we don't agree on it)

AP-protocol.md
franko-fox: I added an extra provision in response to your comment in the latest commit.

franko-fox:
> I really like some of your changes, however there are some possible errors in my opinion:
> 
> What happens if one of the neighbors is not responding with `FloodProcessing` but is also not crashed? In the example below, node `A` receives from `..` the `FloodRequest` (and responds with a `FloodProcessing`), sends the `FloodRequest` to `C` (which responds to it with a `FloodProcessing`). Before `A` can send the `FloodRequest` to `B` as well, `B` receives it from `C`, and then responds to `C` with a `FloodProcessing` and ignores the `FloodRequest` from `A`. At this point, however, `A` knows that not all of its neighbors are done or failed, because `B` is neither, so `A` will block the flooding.
> 
> image

As mentioned in the standard, after these messages are passed:
- A has prev_hop = ... and C as pending, since B hasn't returned a FloodProcessing message it means it received the request from another drone
- C has prev_hop = A and has received a FloodProcessing from B, so it is the only pending drone
- B has prev_hop = C and would've sent a FloodRequest to A, which would get ignored since it has already received it from ..

In this case:
B has no pending nodes for that flood, so would return a flood response to C with [(A,B),(B,C)] in the hashset.
C receives a reply from B, so adds (A,C) and (B,C) (the latter would not get duplicated since it's a set, just ignored) and returns the response to A.
A receives the reply, marks C as done, and adds (..,A) and (A,C) in the connections list, and the return chain keeps going.

This protocol essentially creates a [spanning tree](https://en.wikipedia.org/wiki/Spanning_tree) that contains all nodes in the graph and has its root at the initiator. The flood propagates until the whole tree is formed, in which case the leaves of the tree return replies, and they proceed down all the branches returning to the root. The tree is formed of all the edges where the FloodRequest <-> FloodProcessing messages are exchanged, some edges will be ignored because we only care about **nodes**, since they already know all their neighbors, not all edges.

DaYe03:
> > I really like some of your changes, however there are some possible errors in my opinion:
> > What happens if one of the neighbors is not responding with `FloodProcessing` but is also not crashed? In the example below, node `A` receives from `..` the `FloodRequest` (and responds with a `FloodProcessing`), sends the `FloodRequest` to `C` (which responds to it with a `FloodProcessing`). Before `A` can send the `FloodRequest` to `B` as well, `B` receives it from `C`, and then responds to `C` with a `FloodProcessing` and ignores the `FloodRequest` from `A`. At this point, however, `A` knows that not all of its neighbors are done or failed, because `B` is neither, so `A` will block the flooding.
> > image
> 
> As mentioned in the standard, after these messages are passed:
> 
> * A has prev_hop = ... and C as pending, since B hasn't returned a FloodProcessing message it means it received the request from another drone
> * C has prev_hop = A and has received a FloodProcessing from B, so it is the only pending drone
> * B has prev_hop = C and would've sent a FloodRequest to A, which would get ignored since it has already received it from ..
> 
> In this case: B has no pending nodes for that flood, so would return a flood response to C with [(A,B),(B,C)] in the hashset. C receives a reply from B, so adds (A,C) and (B,C) (the latter would not get duplicated since it's a set, just ignored) and returns the response to A. A receives the reply, marks C as done, and adds (..,A) and (A,C) in the connections list, and the return chain keeps going.
> 
> This protocol essentially creates a [spanning tree](https://en.wikipedia.org/wiki/Spanning_tree) that contains all nodes in the graph and has its root at the initiator. The flood propagates until the whole tree is formed, in which case the leaves of the tree return replies, and they proceed down all the branches returning to the root. The tree is formed of all the edges where the FloodRequest <-> FloodProcessing messages are exchanged, some edges will be ignored because we only care about **nodes**, since they already know all their neighbors, not all edges.

If I understand correctly we need to check if the connections received from a Flood Response contains a pending neighbor and mark it as "done". Is better to clarify this part in the AP-protocol.ms, maybe in line 144: When a **pending** neighbor returns a `FloodData` message, the message is temporarily saved and the node is marked as **done**. When all neighbors are **done** or **failed**, or if there are none (in the case of a node connected to only one neighbor, from which it has received the request) the Node creates its own `FloodData`:

matteo-dinoia:
Voted against.

PR #81, opened 2024-11-25
Remove DroneOptions in new()

StefanoSacchet:
his small pull request includes the removal of the DroneOptions struct and updates to the example usage to reflect these changes. In my opinion it does not make sense to have an extra useless wrapper (as you can see in the changed example).


matteo-dinoia:
Seem fine and make sense

matteo-dinoia:
We can always recreate that option by making a function that take that struct and call the new of the drone

matteo-dinoia:
Please fix the conflict because the example given doesn't work

matteo-dinoia:
You probably merged it wrongly. Maybe creating a PR from zero may be easier

PR #79, opened 2024-11-24
Simulation Controller commands specification 2

DiCesareDaniele:
Duplicate of #72 but Command is split.

```rust
/// From controller to drone
#[derive(Debug, Clone)]
pub enum DroneCommand {
    AddSender(NodeId, Sender<Packet>),
    SetPacketDropRate(f32),
    Crash,
}

/// From drone to controller
#[derive(Debug, Clone)]
pub enum NodeEvent {
    PacketSent(Packet),
    PacketDropped(Packet),
}
```

I have changed the example. It should be better now.
`cargo run --example drone_usage --features="serialize"`
if you want to test it

PR #85, opened 2024-11-25
Added priority for simulation controller messages

franko-fox:
This change adds mandatory priority for simulation controller messages, so there's no risk that the drones might continuously process messages and never pick up on any important messages (like crashes).
Changes:
- Added a new paragraph in the protocol to enforce this priority
- Used select_biased macro in drone example and swapped the two recv(...) so the simulation controller one is at the top


matteo-dinoia (approved):
Fine

DiCesareDaniele:
Deprecated since #79 got merged

franko-fox:
I'm proceeding to close and delete the branch then

DiCesareDaniele:
> I'm proceeding to close and delete the branch then

Can you please reopen this and just change the protocol file? I didn't specify that select biased was forced

franko-fox:
Removed the drone usage edits, and kept only the protocol file change

matteo-dinoia:
> Removed the drone usage edits, and kept only the protocol file change (please ignore cargo fmt not passing)

Is there any problem with cargo fmt?
For the rest, it is ok.

franko-fox:
> > Removed the drone usage edits, and kept only the protocol file change (please ignore cargo fmt not passing)
> 
> Is there any problem with cargo fmt? For the rest, it is ok.

It was complaining about an empty line, I just fixed it.

DiCesareDaniele:
Thank you so much @franko-fox <3. @matteo-dinoia I think now it can be merged

matteo-dinoia:
Did we vote on this? I think not right?

Suge42:
> > Did we vote on this? I think not right?
> > We should probably vote.
> 
> Yes it was in my pr but for some reason it did not got merged. Plus there was an issue about this and everyone who commented said that there was no reason to keep the sim controller trait

If the PR was the one about the Sim commands, in class we voted for the two enums, but not really for the example and the prioritizing of the sim contr commands.
After I've taken a look at all current PR, I'll make a vote for this too on telegram.

Issue #49, opened 2024-11-20
Missing/To be defined


Suge42:
These are the problems that has been recognized and still need to be solved. If you have others, submit them freely and we'll see if it's needed to add them.

If you want to work on any of these, could be useful to declare it, so that too many people don't work on the same stuff, while ignoring other problems.

- Simulation controller commands to drone;
- What happens to channels (and especially messages that could be inside them) when a drone crashes;
- Development of a test suite to check the general functioning of other groups drones after the fair;
- How to create and specify the different drones in initialization;
- Specify minimum connections of clients and drones;
- OPTIONAL: an assembler code to use for client/server;

Suge42:
I started working on the Sim Contr commands

franko-fox:
Relating to what happens to the channels, since the controller knows the topology, it can manually order all the neighboring drones to remove the relative sender channel and send a crash command to the drone. All drones will then stop sending messages, the drone that's about to crash will finish processing in-flight messages, and its own thread will terminate, after a short timeout the controller can send a second crash command which will fail since the channel is disconnected, and can remove everything related to the drone from its memory.

marcoff181:
to recap:
- [X]     Simulation controller commands to drone with #79 
- [X]     What happens to channels (and especially messages that could be inside them) when a drone crashes: after voting #84 we should know what the WG wants to happen  
- [ ]     Development of a test suite to check the general functioning of other groups drones after the fair
- [X]     How to create and specify the different drones in initialization: will be decided with #86 
- [ ]     Specify minimum connections of clients and drones
- [X]     OPTIONAL: an assembler code to use for client/server (I think this is not needed any more because we decided that client/server messages should be up to the group)

matteo-dinoia:
Specify minimum connection has been merged. I don't remember which PR, but we merged it becuase it was written in project main that the network MUST be connected.

matteo-dinoia:
Also the test suite could be a selling point for the drone at the fair

matteo-dinoia:
Didn't we vote to keep everything like it is now? So no broadcast.

PR #86, opened 2024-11-26
Specify drone implementation

marcoff181:
Issue mentioned in #49, I propose a way to initialize the drones without cluttering the initialization file, in my opinion we have no need to specify that a single drone needs to have a particular group implementation. The important bit is that we want the implementations to be equally spread.

There is still the possibility (up to the group's network initializer) of debugging a single drone by filling the network with just their drone. 

It also considers the possibility of not having a 10-drone network, to be able to try all kinds of networks.

If you disagree, and you think we should be able to define in the initialization file the implementation for each drone, please comment or make a PR with a counter proposal so that we can vote on it


matteo-dinoia:
fine

DiCesareDaniele:
Amazing

PR #87, opened 2024-11-26
 Specify minimum connections of clients and drones

marcoff181:
# Issue
there is an issue in #49 that says:
>  Specify minimum connections of clients and drones

I  think this PR is sufficient to clear any doubts about what connections there should be, but we can add to it if there are other bounds on connections between clients and drones.

# Project Main
Just for reference, this is what's in the project main:

> The clients and servers are atthe edge of the network

fixed with this PR

> When starting, each server is connected to at least two drones, and each client is
connected to at most two drones. 

we already have this

> Although the network is unreliable, we assume that network partitions that
separate clients from servers never happen so that it is always possible to
eventually deliver packets.

this PR ensures that at the start of the network this is true, the only other option is a crash


matteo-dinoia:
Fine but i think at this point we have a bit of repetition of the same stuff in different points. Consider condensing other line to simply the protocol.

matteo-dinoia:
Fine, I don't even know if we should vote. Because it was already kinda of implied. What do you think, @Suge42?

Suge42:
> Fine, I don't even know if we should vote. Because it was already kinda of implied. What do you think, @Suge42?

Seems good on its own, I'm just confused by the title since I don't see where we 'specify the minimum connections of clients and drones', but maybe I'm missing something.

marcoff181:
> I'm just confused by the title since I don't see where we 'specify the minimum connections of clients and drones' 

Yeah, but there was an open issue about that, and the only requirement left to define about minimum connections that i could think of was that every node must be connected to the network.

PR #89, opened 2024-11-26
QOL improvements

daw-dev:
This pull request introduces several enhancements and new functionalities to the routing and packet handling components of the network crate. The key changes include the addition of new methods for the `SourceRoutingHeader`, `FloodRequest`, and `Packet` structs, as well as the implementation of the `Debug` trait for better debugging support.

### Enhancements to routing and packet handling:

* [`crates/wg_network/src/routing.rs`](diffhunk://#diff-e9cdda6124f305c9d2db0f6fe698b56b3842b2f7ca4dd0b36f0ee5decd1c7b6cR1-R80): Added multiple methods to `SourceRoutingHeader` for initialization, hop management, and route manipulation. Implemented the `Debug` trait for improved debugging output.

* [`crates/wg_packet/src/flood.rs`](diffhunk://#diff-c0f81269c3732cae7f938a769eb35c55838423e89255fcc6d9072f9062612ee4R17-R34): Added new methods to `FloodRequest` for creation, path trace increment, and response generation.

* [`crates/wg_packet/src/packet.rs`](diffhunk://#diff-afac9b4d5950e1e88c70424af11d4d2b4639afb2250336d64e0325d30757839aR1-R55): Added new constructors to the `Packet` struct for different packet types (fragment, ack, nack, flood request, and flood response). Implemented the `Debug` trait for the `Fragment` struct.

### Suggestions

Please tell me if you would like some things to be implemented differently, if you want something to be added and if you would like to suggest some different namings.


luiss07:
I think that all these things are group contributions, we don't need to have them in the common crate.

daw-dev:
Well, there's a much better (imo) Debug implementation of some structures, that I cannot do inside my group's project and it was to make it look like a real library.
It would be inconvenient if you were to use a library that just gives you some structures with public fields.

I think that if we want to do a fine job with our common library, it would include stuff like these.

daw-dev:
But why shouldn't we create a better implementation for the structs?
I've just created some helper methods (I'm trying to cover every possible need), you can suggest other ones for me to add and worst case scenario you can get immutable and mutable references to the various fields.

luiss07:
> But why shouldn't we create a better implementation for the structs? I've just created some helper methods (I'm trying to cover every possible need), you can suggest other ones for me to add and worst case scenario you can get immutable and mutable references to the various fields.

I understand, but those, IMO, are group contributions and things that you can improve in your code.

daw-dev:
That way we would be forcing every group to either write terrible code (indexing the vec every two lines of code) or make writing better code (which would be really similar for every single group) in a really inconvenient way by wrapping and unwrapping at every function call.
This makes everyone's life easier and the code would be similar for anyone anyway.

daw-dev:
Yeah, and so would my friend in the other group, the other guy of another group and everybody in the class (probably you would do something similar as well, wouldn't you?).
Creating the same code, just less convenient because we have to pass the struct as a parameter or we have to wrap/unwrap every time.
I don't understand your issue with this kind of pull request, if it is something like "I would do things differently", I'll assure you:
1. Stuff will be super generic
    1. No specific-algorithm-related stuff
    2. Just something more than simple getters and setters
3. I'm super down to taking suggestions
4. If you really want something to be different, you can add your own custom functionality 

All this pull request does is turning
```rust
```

DiCesareDaniele:
I am sorry but as @matteo-dinoia and @luiss07 said these are group contributions. Also I would not add functions in general in this repo since this is more to define what gets sent between drones and not functionality

matteo-dinoia:
I think it may be better for you to just do something similar on your own. Thie shouldn't be here.

Suge42:
I don't think closing the pull request was really democratic, since usually with big changes we prefer to vote and eventually deny them.

Also since the PR is a proposal of code, it doesn't make sense to create an issue on it (also because many people don't read issues)

And for last, when we import a library we usually expect to have a bunch of useful methods rather than barebone structs, so I think this proposal should be voted rather than just closed like this.

matteo-dinoia:
Fair, sorry. I guess you are right that nobody read issue.

DiCesareDaniele:
In our code we did something similar to this

```rust
pub mod helper {
    use wg_2024::{
        network::{NodeId, SourceRoutingHeader},
        packet::{Packet, PacketType},
    };

    pub fn get_hop(h: &SourceRoutingHeader) -> Option<NodeId> {
        h.hops.get(h.hop_index).cloned()
    }

    pub fn get_next_hop(h: &SourceRoutingHeader) -> Option<NodeId> {
        h.hops.get(h.hop_index + 1).cloned()
    }
}
```

I would suggest you do the same

matteo-dinoia:
Yeah that was what i was saying

daw-dev:
Yeah yeah, I know, and as I was saying we already have two teams forced to write (excluding mine) the same exact code which is more inconvenient and slightly worse (coding-style wise) than mine. Anyway let's vote on this, if it gets denied, I'll also implement it similarly.

matteo-dinoia:
TODO: Put them back to public.

daw-dev:
I think this is mergeable, if you have some suggestions that need to be made before merging put them here, if you have suggestion about some functions and methods to add either put them here or you can even text me on [telegram](https://t.me/justdaw).

matteo-dinoia:
self.hops.get(0).cloned()
   |         ^^^^^^^^^^^^^^^^ help: try: `self.hops.first()`
   
 Fix this clippy warning pls

matteo-dinoia:
Clippy and Test fails, please fix them

matteo-dinoia:
error[E0308]: mismatched types
   --> crates/wg_packet/src/packet.rs:188:13
    |
188 |             data,
    |             ^^^^ expected an array with a fixed size of 128 elements, found one with 80 elements

error[E0308]: mismatched types
   --> crates/wg_packet/src/packet.rs:199:13
    |
199 |             data,
    |             ^^^^ expected an array with a fixed size of 128 elements, found one with 80 elements

matteo-dinoia:
We can still wait a bit for this.

crates/wg_network/src/routing.rs
```diff
@@ -7,3 +11,182 @@ pub struct SourceRoutingHeader {
...
+                return None;
+            }
+
+            Some(SourceRoutingHeader {
+                hop_index: self.hop_index.max(start) - start,
+                hops: self.hops[start..end].to_vec(),
+            })
+        }
+    }
+}
```
matteo-dinoia, line 192: Isn't there any more compact way to do this?

daw-dev: I know it's weird, if you can figure it out please help me

matteo-dinoia: Maybe use Range https://doc.rust-lang.org/std/ops/struct.Range.html?

matteo-dinoia:
```rust
fn main(){
    let numbers = vec![1, 2, 3, 4, 5, 6, 7, 8, 9];
    let subvec = subvector(&numbers, 2..5);
    println!("{:?}", subvec);
}

fn subvector<T: Clone> (v: &Vec<T>, range: Range<usize>) -> Vec<T>{
    v[range].to_vec()
}
```

matteo-dinoia:
```suggestion
pub fn sub_route(&self, range: impl Range) -> Option<SourceRoutingHeader> {
    Some(SourceRoutingHeader{
        hops: self.self.hops.get(range)?.to_vec(),
        hop_index: self.hop_index.max(range.end - 1) - range.start,
    }
}
```
I put it down fast so i'm not sure it work.

daw-dev: I knew I could do it that way, but the additional complexity is justified by the possibility of doing route.subroute(2..=0) that also reverses the route.
If you think it isn't useful, I'll simplify it.

matteo-dinoia: More than not useful. I honestly hadn't understood that. I was thinking it just checked boundaries.
I think it is not necessary when we have rev. You can always make another function that does both.

crates/wg_network/src/routing.rs
```diff
@@ -1,9 +1,158 @@
...
+    /// Returns the next hop of the route if present.
+    pub fn next_hop(&self) -> Option<NodeId> {
+        self.hops.get(self.hop_index + 1).cloned()
+    }
+    /// Returns the previous hop of the route if present.
+    pub fn previous_hop(&self) -> Option<NodeId> {
+        if self.is_first_hop() {
+            return None;
+        }
+        self.hops.get(self.hop_index - 1).cloned()
```
matteo-dinoia, line 91: Could use ? notation

crates/wg_network/src/routing.rs
```diff
@@ -1,9 +1,158 @@
...
+
+        let end = match range.end_bound() {
+            Bound::Included(&end) => end + 1,
+            Bound::Excluded(&end) => end,
+            Bound::Unbounded => self.hops.len(),
+        };
+
+        Some(SourceRoutingHeader {
+            hop_index: self.hop_index.max(start) - start,
+            hops: self.hops.get(start..end)?.to_vec(),
```
matteo-dinoia, line 150: At this point just getting a range as an input?

matteo-dinoia (commented):
Seem fine. Does the example compile? The github action doesn't test examples.

matteo-dinoia:
Clippy complain for unused variable and peraphs you need to run cargo fmt.

matteo-dinoia:
Let's do it like this: the code still need some improvement i guess, but for now let's get it merged, we will fix it later

daw-dev:
Which kind of improvements?

matteo-dinoia:
Mostly code and performance wise. It could be shorter and easier to understand plus it could be faster.

PR #91, opened 2024-11-26
Add NodeId to Nack::Dropped

di-bon:
This PR adds the NodeId of the drone which dropped the packet to the enum Nack::Dropped. This allows clients and servers to choose packet routes in the network according to the amount of packets dropped by each drone.


daw-dev:
Isn't the NodeId of the drone that dropped the packet just the first hop of the Dropped packet?

di-bon:
Yes, I didn't think about it. Thanks

PR #92, opened 2024-11-26
Removing SimulationController trait

Sm3r:


DiCesareDaniele:
There was likely a problem during the merge of #79. We agreed that we would have removed the Sim Controller trait. Can you please remove the controller file and revert lib.rs. Thanks a lot

Suge42 (approved):

PR #94, opened 2024-11-26
Why remove fragment_index from Nack?

RickyBenevelli:
I did not understand why PR #90 was approved and merged. In my opinion it breaks the protocol and it is necessary to go back to the previous version. 

The first thing I would like to clarify is that in my opinion Nacks can only be created for `MsgFragments`: all other messages cannot be dropped or suffer errors. 

Also, each time a Nack is created, it results in the loss of a `MsgFragment`, so it is necessary to have the ability to report to the host the `fragment_index` to be retransmitted.

### DIFF
--- a/crates/wg_packet/src/packet.rs
+++ b/crates/wg_packet/src/packet.rs
@@ -19,10 +19,16 @@ pub enum PacketType {
 }
 
 #[derive(Debug, Clone)]
-pub enum Nack {
+pub struct Nack {
+    pub fragment_index: u64,
+    pub nack_type: NackType,
+}
+
+#[derive(Debug, Clone)]
+pub enum NackType {
     ErrorInRouting(NodeId), // contains id of not neighbor
     DestinationIsDrone,
-    Dropped(u64), // fragment id of dropped fragment
+    Dropped,
     UnexpectedRecipient(NodeId),
 }
 


daw-dev:
Nack can be created by any other type of packet: for example, if I send a FloodResponse with some route that has a link that doesn't exist, a ErrorInRoute should be sent back, and since the FloodResponse doesn't have a fragment_index it doesn't make any sense to include one in every Nack packet.
It's true that for dropped fragments it's useful to know the referenced fragment_index, that's why the Dropped variant now included a u64 parameter.
If there are more doubts please let's continue this conversation because I know this has been merged quickly.

daw-dev:
Now I see what you mean maybe, fur sure it doesn't make any sense to include the fragment_index for the reasons stated before, maybe we should box the problematic packet all together?
Let me sleep on it.

daw-dev:
Or we could option the fragment index.

RickyBenevelli:
> Also we can assume that if one fragmebt arrived at the wrong destination, then all did.

I think it is better to handle errors well and not make these assumptions. In fact, it is enough that one of the drones I bought is defective for it to occur that one fragment arrives at the correct destination and one at the wrong one. 

In any case, even if we want to make this assumption, handling del nacks with `fragment_index` and nacks without is more complicated than handling them all with `fragment_index`

daw-dev:
No, we cannot assume that, I don't think so.
In order to imply that we would need to also assume that the group's routing algorithm is correct, can we assume this?
Also I think that the bigger problem is if a drone crashes: if I'm sending frag 0 -> 5 and after the 3rd one the drone crashes I want to recover just the last ones.
I think that the better solution could be to Box the problematic packet, so that every drone and every client and server can do whatever they want.
I don't think adding a dummy fragment index is a good solution.

matteo-dinoia:
That  is excessive. And what i mean by we can assume is that if one arrive wrong we can just resend everything. 
If you want to go back to add fragment id that is fine but passing the entire packet is maybe too much

daw-dev:
So if I send 1000 fragments and just one gets routed badly (improbable, but it could happen), I should send back every 1000 of them?

daw-dev:
So we should just create it randomly when we don't have one? That's really bad in my opinion.

matteo-dinoia:
Nono, an Option is fine. I was just against passing back the entire packet

Marui-9:
Could we enclose the box with faulty packet in the NackType? This way we could use it only if it's needed, and avoid it for e.g. DestinationIsDrone nack.

matteo-dinoia:
> Could we enclose the box with faulty packet in the NackType? This way we could use it only if it's needed, and avoid it for e.g. DestinationIsDrone nack.

Sorry what you mean?

RickyBenevelli:
> So we should just create it randomly when we don't have one? That's really bad in my opinion.

I did not understand when it is not useful to insert the fragment_index

daw-dev:
If someway you recieve a FloodResponse that has a path with a link that isn't actually there, you have to send an ErrorInRouting Nack, but you don't have a fragment_index and it's bad practice to generate a dummy one.

We could option it, but then what if i want to recover errors from the floodrequest i send?

For me either:
```rust
fragment_index: Option<u64>
```
 - just a patch
 - cannot handle errors from non-fragment packets
or
```rust
faulty_packet: Box<Packet>
```
 - more complex
 - you can handle whatever you want

I'm team box, but it should be discussed.

daw-dev:
I'd put it in the outer struct, not in the different variants, but yeah.

IlyaEmeliyanov:
I think this pr should be closed because it doesn't follow the protocol specification for only Fragments being dropped. I don't understand why it was modified in such a way that it allows other packets to be dropped. Am I missing something?

daw-dev:
It doesn't allow other packet type do be dropped it allow them to be wrong, what would your drone do if I send a flood response with an invalid path?

RickyBenevelli:
I believe that the only packets for which it may be possible to send a nack must be packets of the `MsgFragment` type.

I think this is something to discuss and decide before voting for this PR. I have opened an Issue #98 to keep the topics separate though. Go ahead and write down which approach you would prefer to take

DiCesareDaniele:
The issue got reverted by #96. Please close this pr

PR #95, opened 2024-11-26
Added limits for simulation controller commands

RickyBenevelli:
In this PR I made explicit something that I think was obvious to everyone, but better to write it down: The simulation controller cannot modify the network making it invalid to the limits we initially imposed

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -423,7 +423,7 @@ The Simulation Controller can execute the following tasks:
 
 The Simulation Controller can send the following commands to drones:
 
-`Crash`: This command makes a drone crash. Upon receiving this command, the drone’s thread should return as soon as possible. 
+`Crash`: This command makes a drone crash. Upon receiving this command, the drone’s thread should return as soon as possible.
 
 `AddSender(dst_id, crossbeam::Sender)`: This command adds `dst_id` to the drone neighbors, with `dst_id` crossbeam::Sender.
 


matteo-dinoia:
Do we want to force them?
Like if we remove and then replug nothing bad should happen. But i guess this is a valid implementation.

DiCesareDaniele:
> In this PR I made explicit something that I think was obvious to everyone, but better to write it down: The simulation controller cannot modify the network making it invalid to the limits we initially imposed

I would not force this at runtime. Just assume this is always the case

matteo-dinoia:
> > In this PR I made explicit something that I think was obvious to everyone, but better to write it down: The simulation controller cannot modify the network making it invalid to the limits we initially imposed
> 
> I would not force this at runtime. Just assume this is always the case

This is the kind of stuff the group can easily decide without causing other problems.

IlyaEmeliyanov:
Seems reasonable to add clarity

Suge42:
> > In this PR I made explicit something that I think was obvious to everyone, but better to write it down: The simulation controller cannot modify the network making it invalid to the limits we initially imposed
> 
> I would not force this at runtime. Just assume this is always the case

Would this mean that the groups have to implement the check inside the simulation controller, or are we just saying that it's assumed that it doesn't happen?

matteo-dinoia:
> > > In this PR I made explicit something that I think was obvious to everyone, but better to write it down: The simulation controller cannot modify the network making it invalid to the limits we initially imposed
> > 
> > I would not force this at runtime. Just assume this is always the case
> 
> Would this mean that the groups have to implement the check inside the simulation controller, or are we just saying that it's assumed that it doesn't happen?

If it is in the protocol it is something we need to respect. 
It is not that hard to implement if we approve it.

matteo-dinoia:
I'm not allowed to fix the conflict because you disable write access to it. So i cannot fix and merge that.

matteo-dinoia:
Please remove the extra space you added (very small nitpick). I'll merge it then.

RickyBenevelli:
> Please remove the extra space you added (very small nitpick). I'll merge it then.

Done

PR #96, opened 2024-11-27
Restore fragment_id

Suge42:
Applying the same changes of PR #94 (check it first), to solve the problems of PR #90, with the addition of a comment to clarify the behavior in case of non-fragment packet's drop.

I know that is the same as the other PR, but whenever the discussion becomes too long not more than 12 WGL read them.

If you have a different approach, it is welcome, but we'll save time just doing another PR and posting a vote on telegram probably since a discussion has already been done on the subject.


matteo-dinoia:
> What if we just call it `offset`? It's more generic and I think it makes more sense (sorry for my acoustic side).

I think offset is worse at communicating what it is.

PR #100, opened 2024-11-28
Add PartialEq

marcoff181:
when doing integration testing for my drone I would like to be able to write 
``` rust
assert_eq!(received_packet,expected_packet)
```
this PR derives (and implements in one case) `PartialEq`  for all structs that are sent with channels: `Packet,NodeEvent,DroneCommand`


DiCesareDaniele:
You wouldn't check if two Fragments are equal by checking their underlying structure but by checking if the triplet (src_id, session_id, fragment_index) are equal. Same for Flood. And adding partialEq to the controller commands seems excessive.

marcoff181:
> You wouldn't check if two Fragments are equal by checking ...

You mean Packets, right? 

 I agree that using partialEq like this would be excessive if we were to use it in the simulation itself, But this is just meant to save time when a group is testing that their Drone can forward packets, flood requests, etc. 

The other option is to leave it to each group to write the logic of comparing Packets. I just thought that, for what it's needed for, this approach would be sufficient.

DiCesareDaniele:
I don't know it's a good feauture for testing but seems kinda forced. I guess we'll vote on this

matteo-dinoia:
Can we do something about having 2 PR about eq/partialeq/... traits?

DiCesareDaniele:
> Can we do something about having 2 PR about eq/partialeq/... traits?

I would approve mine since it just adds Eq/PartialEq for trivial types and I kinda need it to develop 😂. And do a vote on this on Monday

matteo-dinoia:
Seem fine. I'll merge yours them, because that one it has nothing that people could be against.

matteo-dinoia:
TODO: make this an optional feature.

PR #106, opened 2024-11-29
Data size in Fragment

RickyBenevelli:
I changed the size of the data array to be a power of 2: from 80 to 128


DiCesareDaniele:
> I changed the size of the data array to be a power of 2: from 80 to 128

Wholesome. You think we can extract the size into a const?

Like this
```rust
pub const FRAGMENT_DSIZE: usize = 128;

pub struct Fragment {
    pub fragment_index: u64,
    pub total_n_fragments: u64,
    pub length: u8,
    pub data: [u8; FRAGMENT_DSIZE],
}
```

franko-fox:
Why not make the fragment size 1 kilobyte (1024 bytes) or a multiple of that? Since we need to send media through it.

Suge42:
> Why not make the fragment size 1 kilobyte (1024 bytes) or a multiple of that? Since we need to send media through it.

Since also text messages are fragmented, it might be too big, and just send whole messages, which would make the use of fragments useless in that scenario. Sticking to 128 is probably better, but could be discussed.

franko-fox:
> > Why not make the fragment size 1 kilobyte (1024 bytes) or a multiple of that? Since we need to send media through it.
> 
> Since also text messages are fragmented, it might be too big, and just send whole messages, which would make the use of fragments useless in that scenario. Sticking to 128 is probably better, but could be discussed.

Issue with that is that Media will take *forever* to send, since it needs to be split into tiny fragments.

matteo-dinoia:
Please fix merge conflict as i have no write access to your branch since you disabled

PR #107, opened 2024-11-30
Test cases

luiss07:
In this PR we propose some test cases that the whole WG could use to show that your drone is working during the drone fair. 

As of today, this PR contains only the tests for the `Packets`, mainly for two reasons:
- We are working on the other ones.
- We'd like to have feedback on whether you think this could be useful, thus we will continue to pull tests (that we can manage to write before the freeze) or not.

The idea of this code is to have some test functions that do not depend on the drone implementation and can be used by anyone. Just to be as exhaustive as possible below you can check our test implementation for the `generic_packet_forward` function.

```rust
use crossbeam::channel::unbounded;
use rusteze_drone::RustezeDrone;
use std::collections::HashMap;
use wg_internal::drone::{Drone, DroneOptions};
use wg_internal::packet::Packet;
use wg_internal::tests::*;

#[test]
fn packet_forward() {
    // drone 11
    let (d_send, d_recv) = unbounded();
    // drone 12
    let (d2_send, d2_recv) = unbounded::<Packet>();

    let (_d_command_send, d_command_recv) = unbounded();

    let neighbours = HashMap::from([(12, d2_send.clone())]);
    let mut drone = RustezeDrone::new(DroneOptions {
        id: 11,
        pdr: 0.0,
        packet_send: neighbours,
        packet_recv: d_recv.clone(),
        controller_send: unbounded().0,
        controller_recv: d_command_recv,
    });

    generic_packet_forward(drone, &d_send, &d2_recv);
}
```

I was thinking of adding the code above as an example but since it is using our implementation of the drone I am not sure where to put it.

> I also added the `PartialEq` for the Packet struct (and sub-struct). I didn't really want to implement a custom (and messy) one to check if two Packets are equal. I know there is already an open PR about this, but it is needed to make the `assert_eq!` work.

Let me know what do you think about it.


matteo-dinoia:
This should be probably part of the group project. As the professor stated is one of the most important distinguishing factor.
Also being part of the drone we can use the native way to create test.

luiss07:
Yep, however, I was also thinking that if each of us has its own tests then we don't have any way of measurement to state that a drone is working (or at least has some functionalities) before buying it. I know that it might be the "fun part of the fair" but I'd like to have something concrete rather than just trusting words. 

That's why I don't know if this might be useful.

Since I just noticed that we will have two WG next week, we might wait to see what the people say on Monday and eventually pull other tests on Tuesday, or just close the PR.

matteo-dinoia:
The idea is to only buy software with various tests. You can see the tests and evaluate them i guess.

DiCesareDaniele:
Nice but I would have initialized the drones in the tests like this
```rust
fn generic_test<T: Drone>() {
    let drone = T::new(...);
}
```
so there is less trouble using it. 
Good starting point, but not ready 👍

luiss07:
I pushed the drone initialization within the test cases. If you want to wait until you merge the DroneOption (and PartialEq), then I can edit it accordingly and do another push without the DroneOptions.

matteo-dinoia:
> I pushed the drone initialization within the test cases. If you want to wait until you merge the DroneOption (and PartialEq), then I can edit it accordingly and do another push without the DroneOptions.

Done, please updated this PR so we can get it merged

PR #109, opened 2024-12-01
Ack, Nack and FloodResponse through crashed drones

Suge42:
Since 'Ack' and 'Nack' can't be lost, in case of an error during their routing, a way to reassure their arrival to destination is to send them there through the simulation controller.

I also put 'FloodResponse's because I thought we might want to receive them too, since otherwise we would risk to not have a complete knowledge of the network, and any error could be corrected after.
If the WGL thinks that's excessive, we can vote to keep just 'Ack's and 'Nack's


matteo-dinoia:
Let's merge all the voted change about crashing node in a single PR.

crates/wg_drone/src/drone.rs
```diff
@@ -10,6 +10,7 @@ pub struct DroneOptions {
     pub id: NodeId,
     pub controller_send: Sender<NodeEvent>,
     pub controller_recv: Receiver<DroneCommand>,
+    pub controller_shortcut: Sender<Packet>,
```
matteo-dinoia, line 13: Should be a node event

matteo-dinoia:
Closed in favor of #114

PR #113, opened 2024-12-02
Clarification of SC - Nodes commands

Isax03:
This PR updates the documentation to address a potential source of confusion in the Simulation Controller (SC) command protocol. Specifically, the SC commands and events have been restricted to drones, as they were mistakenly described as applicable to all nodes (including clients and servers). The changes reflect the following rationale:

1. **Unnecessary Overhead for Clients and Servers**:
   - Most commands (`PacketDropped`, `Crash`, `SetPacketDropRate`) are only relevant to drones and serve no purpose for clients or servers.
   - This avoids imposing unnecessary constraints or requiring additional attributes for clients and servers to communicate with the SC.

2. **Customizability**:
   - `AddSender(dst_id, crossbeam::Sender)` and `PacketSent(packet)` are the only potentially useful commands for clients and servers. Groups can now define their own commands for hosts (clients/servers) in separate, custom enums tailored to their needs.

>  Note: I also removed an unnecessary and nonsensical section, probably the result of an erroneous copy/paste, as I didn’t want to create two separate PRs for something trivial.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -400,7 +400,7 @@ Suppose that client A wants to send a message to server D.
 # Simulation Controller
 
 Like nodes, the **Simulation Controller** (SC) runs on a thread. It must retain a means of communication with all nodes of the network, even when drones go down. 
-The Simulation controller can send and receive different commands to/from the nodes (drones, clients and servers) through reserved channels. The list of available **commands** is as follows:
+The Simulation controller can send and receive different commands to/from the drones through reserved channels. The list of available **commands** is as follows:
 
 ```rust
 /// From controller to drone
@@ -411,10 +411,13 @@ pub enum DroneCommand {
 }
 
 /// From drone to controller
-pub enum NodeEvent {
+pub enum DroneEvent {
     PacketSent(Packet),
     PacketDropped(Packet),
 }
+
+// For backward compatibility with the previous naming
+type NodeEvent = DroneEvent;
 ```
 
 The Simulation Controller can execute the following tasks:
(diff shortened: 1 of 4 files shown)


RickyBenevelli:
good fix

matteo-dinoia:
AddSender(dst_id, crossbeam::Sender) and PacketSent(packet)
But we also need RemoveSender

Isax03:
> But we also need RemoveSender

Yeah, now I'll add, I don't know why there isn't that command

---
> So we would need 4 enums.

I don't understand... What do you mean? In the common crate we only need 2 enums: `DroneCommands` and `DroneEvent`. Other enums are at the discretion of the group.

---
> Also (maybe not yet, wait still) you need to change also the code.

Sure, I forgot to change it

matteo-dinoia:
4 enums as in:
1. type to be sent SC to Drone (+viceversa)
2. type to be sent SC to Client/Server (+viceversa)

I'm not saying it is the best way to do it. But that is what you are proposining in a way.

Isax03:
Yes, but implementing the second point in the common crate isn't necessary. The interaction between SC and Client or Server can vary depending on each group's specific implementation

matteo-dinoia:
> Yes, but implementing the second point in the common crate isn't necessary. The interaction between SC and Client or Server can vary depending on each group's specific implementation

No? They should be standard and decided now. The SC does not communicate in any strange way to the client, like it shouldn't ask the client to send a message. We really should ask the professor how the client work.

Isax03:
No. If a group decides to share specific information with the SC, such as retrieving periodic values from clients and displaying them on a UI, this could involve custom request/response channels with a custom enum. In such cases, the SC-client and SC-server interactions can be tailored to each group's needs

PR #108, opened 2024-12-01
Management of messages inside the channels at crash

Suge42:
Since our implementation of the network doesn't use timers, but we instead use Acks and Nacks, in the case a drone would crash we would still need a way to manage eventual incoming packets to it.

With this method we would stop every other message to arrive to the drone, and then process everything left before crashing as soon as possible.

Also quickly specified how the crashing drone would treat the different types of packet.

PR #114, opened 2024-12-02
Managing Drone crash (union of PR #108  and #109)

Suge42:


crates/wg_controller/src/command.rs
```diff
@@ -32,4 +33,5 @@ impl PartialEq for DroneCommand {
 pub enum NodeEvent {
     PacketSent(Packet),
     PacketDropped(Packet),
+    ControllerShortcut(Packet), //Used for direct routing of Ack, Nack and FloodResponse
```
matteo-dinoia, line 36: Modify this also in the Protocol (you forgot for NodeEvent only)

AP-protocol.md
```diff
@@ -454,6 +472,9 @@ Due to the importance of these messages, drones MUST prioritize handling command
 
 This can be done by using [the select_biased! macro](https://shadow.github.io/docs/rust/crossbeam/channel/macro.select_biased.html) and putting the simulation controller channel first, as seen in the example.
 
+## Shortcut for Ack and Nack
```
matteo-dinoia, line 475: Also for FloodResponse

matteo-dinoia (commented):
Pretty good. The only thing is that we are repeating the same in thing in multiple part and sometimes it even change slightly.

matteo-dinoia:
Small clarification we may want to write down. After receiving a crash message a drone want to call `recv()` on its `Receiver<Packet>` and stop listening to simulation controller. If this doesn't happen you will wait forever for the channel between SC and drone to be closed (which will never happen)

matteo-dinoia:
Ok i'll merge it

Issue #118, opened 2024-12-04
Why are the test not testing


giorebecchi:
I tried to test my drone functionalities using the common tests, but (correct me if I'm wrong) the test called "test_generic_chain_fragment_drop()" and "test_generic_chain_fragment_ack()" are expecting each drone to send back an Ack whenever they don't lose their packet. In the AP_Protocol is written something else... (i.e. only clients and server are responsible to send Acks backwards). Am I missing something?

PR #121, opened 2024-12-07
Fixed test cases

luiss07:
I updated the tests according to the protocol as reported in #118.

Additionally, I added the channels to handle DroneEvents which were uninitialized and could have resulted in the tests not working (as pointed out on telegram).

Let me know if you think the fixes seem reasonable.

> Note: I had to add the wg_controller as the test dependency since it was not used before. And some other minor adjustments.

### DIFF
--- a/crates/wg_tests/src/test_fragments.rs
+++ b/crates/wg_tests/src/test_fragments.rs
@@ -61,7 +61,10 @@ pub fn generic_fragment_forward<T: Drone + Send + 'static>() {
     // d2 receives packet from d1
     assert_eq!(d2_recv.recv_timeout(TIMEOUT).unwrap(), msg);
     // SC listen for event from the drone
-    assert_eq!(d_event_recv.recv_timeout(TIMEOUT).unwrap(), DroneEvent::PacketSent(msg));
+    assert_eq!(
+        d_event_recv.recv_timeout(TIMEOUT).unwrap(),
+        DroneEvent::PacketSent(msg)
+    );
 }
 
 /// Checks if the packet is dropped by one drone. The assert consists in checking if the "client" and "SC" receive the correct packet.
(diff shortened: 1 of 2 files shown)


matteo-dinoia:
Also change all recv to recv_timeout. Or the test will take 5 second each to fail. 
The rest of the test, i'll look later. But i think we are trying to solve the problem of a broken protocol with test that acts as protocols.

luiss07:
> Also change all recv to recv_timeout. Or the test will take 5 second each to fail. The rest of the test, i'll look later. But i think we are trying to solve the problem of a broken protocol with test that acts as protocols.

Do you mean that you want me to put less than 5 seconds? I don't get the second sentence

crates/wg_tests/src/test_fragments.rs
```diff
@@ -86,25 +92,29 @@ pub fn generic_fragment_drop<T: Drone + Send + 'static>() {
...
-    };
-    let nack_packet = Packet {
-        pack_type: PacketType::Nack(dropped),
-        routing_header: srh,
-        session_id: 1,
-    };
+    let nack_packet = Packet::new_nack(
+        SourceRoutingHeader {
+            hop_index: 1,
+            hops: vec![11, 1],
```
matteo-dinoia, line 98: Should be 12, 11 1

matteo-dinoia:
> > Also change all recv to recv_timeout. Or the test will take 5 second each to fail. The rest of the test, i'll look later. But i think we are trying to solve the problem of a broken protocol with test that acts as protocols.
> 
> Do you mean that you want me to put less than 5 seconds? I don't get the second sentence

Yeah like 300 ms is way more than enough. 
The second is about the fact we are trying to fix edge case by creating a single test suite instead of a working protocol.

crates/wg_tests/src/test_fragments.rs
luiss07: Are you sure? In this case the packet is being dropped by 11, it did not reach 12

crates/wg_tests/src/test_fragments.rs
```diff
@@ -86,25 +92,29 @@ pub fn generic_fragment_drop<T: Drone + Send + 'static>() {
...
+            nack_type: NackType::Dropped,
+        },
+    );
 
     // Client listens for packet from the drone (Dropped Nack)
     assert_eq!(c_recv.recv().unwrap(), nack_packet);
+    // SC listen for event from the drone
+    assert_eq!(
+        d_event_recv.recv().unwrap(),
+        DroneEvent::PacketDropped(nack_packet)
```
matteo-dinoia, line 112: It should be the packet dropped not the nack. We should also decide when sending the nack we should also send a drone event of packetSend

matteo-dinoia:
In general, just the approach we are taking but at the moment let's just merge this PR.

luiss07:
> In general, just the approach we are taking but at the moment let's just merge this PR.

Added the `recv_timeout`. The last thing to verify is whether the the Nack that you pointed out in the unresolved conversation should contain 12 or not.

matteo-dinoia:
Yeah sorry i got it wrongly. Merging.

PR #122, opened 2024-12-18
changed new definition to allow for trait objects

ZGako:


Sm3r:
I'm writing on behalf of @Matteo-di-noia, since is GH has been flagged.
"Should be fine. Can we merge it @Suge42? Also we probably also want that Drone to be a subtrait of Send. Both of which shouldn't cause problem."

Suge42:
> Also we probably also want that Drone to be a subtrait of Send. Both of which shouldn't cause problem.

If it doesn't cause any problem to existing drones, it's fine I'd say, change that and if this doesn't break anything we can pull.

PR #124, opened 2024-12-27
Adding Hash trait for NodeType

Danylo37:


din-oia:
Closing because talking to the proposer we solved the issue in a better way.
Using HashMap<NodeId, NodeType> instead of things like HashSet<(NodeId,NodeType)>

din-oia:
Closing because talking to the proposer we solved the issue in a better way.
Using HashMap<NodeId, NodeType> instead of things like HashSet<(NodeId,NodeType)>

PR #125, opened 2024-12-28
Adding NodeId to Dropped Nack

Suge42:
In the protocol is said that drones can have 100% probability and that we should take that into account and find solutions for it.
For some reason we forgot to let clients and servers know which drone dropped the packet, which makes trying to give a weight to the nodes difficult and/or inefficient.

If there's a solution that I missed in the common code, let me know.

Will need all the groups to do some fix inside their network, but in the drones should be really fast and anywhere else if you didn't weight the nodes you can just ignore it.

### DIFF
--- a/crates/wg_packet/src/packet.rs
+++ b/crates/wg_packet/src/packet.rs
@@ -116,7 +116,7 @@ pub struct Nack {
 pub enum NackType {
     ErrorInRouting(NodeId), // contains id of not neighbor
     DestinationIsDrone,
-    Dropped,
+    Dropped(NodeId), // id of the drone that's dropping
     UnexpectedRecipient(NodeId),
 }
 


din-oia:
It is not needed as because you can check which drone originated the Nack by simply checking the RoutingHeader. Let me know if this solve the issue.

Suge42:
> It is not needed as because you can check which drone originated the Nack by simply checking the RoutingHeader. Let me know if this solve the issue.

In the protocol it's actually stated that the Source Routing Headed is obtained by reversing the one of the original message.
It doesn't say anywhere that it should be cut at dropping node.

luiss07:
> > It is not needed as because you can check which drone originated the Nack by simply checking the RoutingHeader. Let me know if this solve the issue.
> 
> In the protocol it's actually stated that the Source Routing Headed is obtained by reversing the one of the original message. It doesn't say anywhere that it should be cut at dropping node.

It is not stated, but out of 10 drones that we bought all of them do it.  Otherwise the drones should handle both cases, which changes how the hop_index system works, and i dont think it is necessary. Thus if some of your drones does not cot the SRH I suggest you ask them to adapt.

daw-dev:
I think that logically the node that dropped the fragment is the source of the `Dropped` packet.
The `Dropped` packet is a new packet from the drone that dropped (the source of the `Dropped` packet) towards the node that originally sent it (the destination of the `Dropped` packet).
The protocol says to reverse the route just to answer the question "how can a drone know how to send back the packet if it has no knowledge of the topology of the network?".
It's true that it isn't specified in the protocol and it should, but I think that this is the obvious solution.

Suge42:
Okok fair enough

PR #126, opened 2024-12-31
PUP1: using pair (flood_id, initiation_id) to see if it was already been received

ghost:
At the moment we should use only `flood_id` to see if we already seen a flood. Because multiple drone could use the same flood_id it we need to use also `initiatior_id`. 

This is what most (but probably not all drone) already do.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -191,11 +191,11 @@ struct FloodRequest {
 ### **Neighbor Response**
 
 When a neighbor node receives the flood request, it processes it based on the following rules:
-- If the flood ID has already been received:
+- If the pair (flood_id, initiator_id) has already been received:
 	- The drone adds itself to the `path_trace`.
 	- The drone creates a `FloodResponse` and sends it back.
 
-- If the flood ID has not yet been received:
+- If the pair (flood_id, initiator_id) has not yet been received:
 	- The drone adds itself to the `path_trace`.
 	- **If it has neighbors** (excluding the one from which it received the `FloodRequest`):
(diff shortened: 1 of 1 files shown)


daw-dev:
It isn't well specified, but as I interpret it, the protocol says that we have to use the `flood_id` to check which one we already got, it never says to ONLY use it (I know it's actually not like this, but this interpretation at least makes sense).
Also, if the flood_id was randomly generated we would have almost zero probability to pick the same and if we did I think a well enough developed Client/Server would recover from the unexpected behaviour.

din-oia:
> It isn't well specified, but as I interpret it, the protocol says that we have to use the `flood_id` to check which one we already got, it never says to ONLY use it (I know it's actually not like this, but this interpretation at least makes sense). Also, if the `flood_id` was randomly generated we would have almost zero probability to pick the same and if we did I think a well enough developed Client/Server would recover from the unexpected behaviour. I actually think it would be cool to see.

It specify to check `flood_id` if already decided so that make the check of `initiator_id` not conformant to the current specs. The protocols state what to do if it is present or not, so checking another value and deciding also because of this (which most drone do at the moment) is against current protocol.
The second point you said doesn't matter as most drone already ignore the protocol.

daw-dev:
As I like to see it, the protocol is like the law and everything we're saying after the freeze of the protocol are like the smaller regulations that explain how to adhere to the protocol well.
Anyway, what if we make the flood_id randomly generated and actually check only the flood_id so that clients and servers need to correctly recover from the (improbable) errors? Should I make another PR about this? I'm probably the only one interested to see this 🙃

daw-dev:
Anyway, I made the PR #133 about this!

PR #127, opened 2024-12-31
PUP2: Clarify how path_trace and its uses.

ghost:
At the moment most drone implementation assumes that a client that initialize a `FloodRequest` also put itself, but that is not necessarily required and as such even an implementation that does not assume so is okay (by using the `initiator_id` field when reconstructing the `FloodResponse`). 

This two implementation can have problem when together and as such we should decide on a single of this two methods (in this PR we use the most common of assuming the client also send itself in the `path_trace`). 

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -188,6 +188,13 @@ struct FloodRequest {
 }
 ```
 
+**NOTE**: that the the FloodRequest should be initializated with the path containing the node that has originated the request, as follows:
+
+```rust
+let req = FloodRequest {flood_id: <some_id>, initiator_id: <this_node_id>,
+	path: vec![(<this_node_id>, <this_node_type>)]}
+```
+
 ### **Neighbor Response**
 
 When a neighbor node receives the flood request, it processes it based on the following rules:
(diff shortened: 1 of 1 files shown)


daw-dev:
I don't think this should be in the common protocol, the `generate_response` method already adds the `initiator_id` at the end of the route of the response only if the first element of the `path_trace` is a client or a server. And even if you don't use the common api methods, this is a problem left to the group to be correctly solved by the drone implementation (I kind of spoiled a possible solution already) because I don't think that it breaks anything.
If I'm not seeing something, please tell me.

Isax03:
I think the issue should be handled by each group individually. If a group bought a drone that assumes the client (or server) ID is already in the path trace, then whoever develops the client (or server) must ensure their own ID is added to the path trace to make it work.

Each group offered a specific implementation and decided which ones to purchase. Therefore, it is now up to each group to integrate all the implementations so that they work together. Alternatively, groups could be contacted one by one to find common ground on critical points.

But I don't think a common spec is needed

din-oia:
> I think the issue should be handled by each group individually. If a group bought a drone that assumes the client (or server) ID is already in the path trace, then whoever develops the client (or server) must ensure their own ID is added to the path trace to make it work.
> 
> Each group offered a specific implementation and decided which ones to purchase. Therefore, it is now up to each group to integrate all the implementations so that they work together. Alternatively, groups could be contacted one by one to find common ground on critical points.
> 
> But I don't think a common spec is needed

Remember that a protocol should allow, if followed correctly communication/working with other following the protocols. At the moment that is not a grantee. As creating a working protocol is an important part of the project we should strive to do so.
Workarounds and especially between some groups decision to integrate should be avoided.

daw-dev:
As I said, drones can easily work with BOTH version just using the common method, we don't have to specify anything IMO

Isax03:
This issue is present only in response generation (if I remember correctly). So other drones have not to know how the response is built or if the client is present or not. They have only to forward the response. Therefore the communication/working with other isn't affected by the specification (or not)

daw-dev:
Yeah but to send the response they have to create the route by reversing the path_trace, so if the initiator added itself in the path_trace they just have to reverse it, if not they also have to push the initiator_id to the hops vector. I think drones can be implemented to just do both so that groups can do whatever they want.

PR #128, opened 2024-12-31
PUP3: clarify floodrequest handling

ghost:
Mostly a clarification, because some drone ignore this part of the protocol.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -211,7 +211,6 @@ struct FloodResponse {
 
 #### Notes:
 - For the discovery protocol, `Packet`s of type `FloodRequest` and `FloodResponse` will be sent.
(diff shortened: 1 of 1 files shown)


daw-dev:
This wasn't previously well specified in the protocol, I agree, but I think it was implied by the ordering of the specification: we firstly explain how to handle FloodRequests, then how to handle other packets. A "exluding FloodRequests" would be the right thing to do, but I wouldn't unfreeze the protocol for this.

din-oia:
> This wasn't previously well specified in the protocol, I agree, but I think it was implied by the ordering of the specification: we firstly explain how to handle FloodRequests, then how to handle other packets. A "exluding FloodRequests" would be the right thing to do, but I wouldn't unfreeze the protocol for this.

Remember the unfreezing is for any number of change (so the unfreezing is for the other more important issue, and while at it we can clarify other parts.

daw-dev:
Yeah, of course if the protocol were to be unfreezed I'd accept this PR

PR #129, opened 2024-12-31
PUP4: fix handling of error of non droppable packets

ghost:
In this PR:
* Clarify how to create the Nack wich is required for some error (like `Dropped` but it was currently not specified)
* Clarify what to do in case of error (both when a packet can be "dropped" and when a packet cannot be "dropped")
* Fix an infinite loop if the drone receive a packet not "droppable" with it as destination.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -324,6 +324,7 @@ Once that the client or server has received all fragments (that is, `fragment_in
 
 # Drone Protocol
 
+## Normal packet handling (excluding FloodRequest)
 When a drone receives a packet, it **must** perform the following steps:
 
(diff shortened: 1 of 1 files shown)


daw-dev:
I didn't understand this PR, everything I read was already in the protocol I think, or at least it was implied by other stuff.

din-oia:
Maybe the name is confusing: it is a rewrite and the most important (as for the protocol change):
1. is the avoiding of creation of a destinationIsDrone when a nack/ack is received with destination that drone that would cause the drone to loop forever.
2. Strictly specified how to handle error in case of Ack/Nack/FloodResponse

daw-dev:
For 1. I think drones should be sending the DestinationIsDrone Nack only in response of fragments and 2. is already specified well enough in the protocol

PR #130, opened 2024-12-31
PUP5: define which error in case of crashing drone

ghost:
Clarify how to return error in case of drone crashing.


daw-dev:
The protocol says:
> - At the same time the Crash command will be sent to the drone, which will put it in 'Crashing behavior'. In this state the drone will call the 'recv()' function only on its 'Receiver' channel, process the remaining messages as follows, then when all the sender to it's channel will be removed, and the channel will be emptied, trying to listen to it will give back an error, which will mean that the drone can finally crash.
> - While in this state, the drone will process the remaining messages as follows:
>   - FloodRequest can be lost during the process.
>   - Ack, Nack and FloodResponse should still be forwarded to the next hop.
>   - Other types of packets will send an 'ErrorInRouting' Nack back, since the drone has crashed.

Why isn't this enough?

PR #131, opened 2025-01-01
PUP6: sequential flood_id

ghost:
As it is now to memorize the previous flood_id it requires the use of a HashMap that contains a single entry for each flood in the whole network. Putting it as sequential would greatly reduce memory usage, and possibly avoid OOM.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -188,6 +188,8 @@ struct FloodRequest {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


daw-dev:
I'd leave it like it is, there are other ways to improve performance IMO. Also, I don't think our simulations will be big enough to let this be an issue.
For example we could have our drones remove just memorize the latest `flood_id` for each `initiator_id`, so every drone would just have just a couple of ids.

daw-dev:
Also, I'd put the extra label in this PR.

din-oia:
> Also, I'd put the extra label in this PR.

You should be able to do that I believe (I added it).

PR #132, opened 2025-01-01
PUP 7: Adding a way to kill drones threads for testing cleanup.

ghost:
At the moment it is not possible so integration testing accumulate a lot of thread that run with no reason.

At the moment the new command only work if crash was not called before we can make it work even if the Crash was called.

### DIFF
--- a/AP-protocol.md
+++ b/AP-protocol.md
@@ -420,6 +420,7 @@ pub enum DroneCommand {
     AddSender(NodeId, Sender<Packet>),
     SetPacketDropRate(f32),
     Crash,
+    KillThread,
 }
 
 /// From drone to controller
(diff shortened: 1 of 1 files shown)


daw-dev:
In my [testing library](https://github.com/daw-dev/drone-tester) i just send the `RemoveChannel` and `Crash` commands, why shouldn't it work?

din-oia:
> In my [testing library](https://github.com/daw-dev/drone-tester) i just send the `RemoveChannel` and `Crash` commands, why shouldn't it work?

It doesn't on most drone impl, I have already tested. Also when closing everything you don't care for the drone to terminate sending to next drone as all drone will die. That even if it would work would cause slow down.

daw-dev:
If it doesn't work you should tell the group, they should automatically shut down with the crash behaviour. There's a little overhead, but not much more than the new command I think.

din-oia:
You would be right if not almost all drones (except us and yours) don't follow that, and will never terminate.

daw-dev:
I was so scared you would say ours did too... Anyway I don't think this PR would solve the problem, you'll just have to break the maroons (if you know what I mean) to the other groups.

din-oia:
> I was so scared you would say ours did too... Anyway I don't think this PR would solve the problem, you'll just have to break the maroons (if you know what I mean) to the other groups.

Partially but it is not the right solution. To test you would actually want this. Also this is an extra so will only be insert if other are voted).

PR #135, opened 2025-01-05
Small `FloodResponse` generation fix

daw-dev:
There was a little problem in the code that I didn't notice when developing it: if the `FloodRequest` has an empty `path_trace` (i.e. if it is the drone directly attached to the initiator and the initiator didn't add itself in the `path_trace`) the method didn't generate a correct `routing_header` for the `FloodResponse`.
This problem shouldn't occur in a simulation, but just to be precise I think we can include this small change in the code.


matteo-dinoia:
I guess it could be merged but it doesn't change much.

matteo-dinoia:
Closing as we clearly don't have time to merge it

